PR #40, opened 2020-10-05
add currency conversion functions and address conversion functions

nud3l:


src/utils/bitcoin.ts
```diff
@@ -0,0 +1,16 @@
+import * as bitcoin from "bitcoinjs-lib";
+
+export function getP2WPKHFromH160(hash: string): string | undefined {
```
gregdhill, line 3:
```suggestion
export function getP2WPKHFromH160(hash: Buffer): string | undefined {
```

src/utils/bitcoin.ts
```diff
@@ -0,0 +1,16 @@
...
+    let btcAddress: string | undefined;
+    try {
+        // TODO: specify script format in parachain
+        const payment = bitcoin.payments.p2wpkh({
+            hash: Buffer.from(hash),
```
gregdhill, line 8:
```suggestion
            hash
```

PR #43, opened 2020-10-06
Execute issue promise

savudani8:


src/apis/issue.ts
```diff
@@ -38,7 +38,28 @@ export class DefaultIssueAPI implements IssueAPI {
     }
 
     private getIssueIdFromEvents(events: EventRecord[]): Hash {
-        return this.api.createType("Hash", events[2].event.data[0]);
+        if (events.length == 4) {
```
danhper, line 41: Maybe add a comment about this `4` here or make it a constant?

src/apis/issue.ts
```diff
@@ -38,7 +38,28 @@ export class DefaultIssueAPI implements IssueAPI {
...
+            const hash = this.api.createType("Hash", events[2].event.data[0]);
```
danhper, line 42: Same here, add a note about why `2`

src/apis/issue.ts
```diff
@@ -38,7 +38,28 @@ export class DefaultIssueAPI implements IssueAPI {
...
+        // - vaultRegistry.IssueTokens
+        // - system.NewAccount
+        // - polkaBtc.Endowed
+        // - treasury.Mint
+        // - issue.ExecuteIssue
+        // - system.ExtrinsicSuccess
+
+        // Assuming `system.NewAccount` only occurs on
+        // first execution, at least 5 events are emitted
+        if (events.length > 5) {
```
danhper, line 59: If at least 5 events are needed, shouldn't it be `>= 5`?

danhper (changes requested):
A few comments here and there could help a bit. Please also check that `isExecutionSucessful` is behaving as expected, I'm not entirely sure. After that I think it should be good to go!

danhper (approved):
Looks good!

PR #45, opened 2020-10-06
Add pagination to listings

savudani8:
Uses iterators to fetch listings in batches of specified size. It adds the `pagedIterator` generator function to `utils`

### DIFF
--- a/src/utils/pagedIterator.ts
+++ b/src/utils/pagedIterator.ts
@@ -2,39 +2,22 @@ import { Issue as IssueRequest, Vault, Redeem, ActiveStakedRelayer } from "../in
 import { AugmentedQuery, PaginationOptions } from "@polkadot/api/types";
 import { StorageKey } from "@polkadot/types/primitive";
 import { Codec, Observable, CodecArg } from "@polkadot/types/types";
-import { ApiPromise } from "@polkadot/api";
 import { QueryableStorageEntry, ApiTypes } from "@polkadot/api/types";
 
 export type PolkadotCodecType = IssueRequest | Redeem | Vault | ActiveStakedRelayer;
 
 export async function* pagedIterator<T extends PolkadotCodecType>(
-    api: ApiPromise,
     polkadotListings: AugmentedQuery<ApiTypes, (arg: CodecArg) => Observable<T>> & QueryableStorageEntry<ApiTypes>,
     perPage: number
 ): AsyncGenerator<T[]> {
-    let startKey: StorageKey = api.createType("Bytes", "") as StorageKey;
+    let startKey = undefined;
     let issueRequests: [StorageKey, Codec][];
     while (true) {
-        // a blank StorageKey is 0x in hex
-        if (startKey.toHex() == "0x") {
-            if (!isRemainingListNonempty(polkadotListings)) {
-                return undefined;
-            }
-            issueRequests = await listPaged(polkadotListings, perPage);
-        } else {
-            if (!isRemainingListNonempty(polkadotListings, startKey)) {
-                return undefined;
-            }
-            const keys = await api.query.issue.issueRequests.keysPaged(
-                { pageSize: perPage, startKey: startKey.toHex() } as PaginationOptions
-            );
-            if (keys.length) {
-                issueRequests = await listPaged(polkadotListings, perPage, startKey);
-            } else {
-                return undefined;
-            }
+        perPage = await nextPageLength(polkadotListings, perPage, startKey);
+        if (!perPage) {
+            break;
         }
-
+        issueRequests = await listPaged(polkadotListings, perPage, startKey);
         startKey = issueRequests[issueRequests.length - 1][0];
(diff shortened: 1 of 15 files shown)


src/utils/pagedIterator.ts
```diff
@@ -0,0 +1,85 @@
...
+    while (true) {
+        // a blank StorageKey is 0x in hex
+        if (startKey.toHex() == "0x") {
+            if (!isRemainingListNonempty(polkadotListings)) {
+                return undefined;
+            }
+            issueRequests = await listPaged(polkadotListings, perPage);
+        } else {
+            if (!isRemainingListNonempty(polkadotListings, startKey)) {
+                return undefined;
```
danhper, line 26: minor: I would just break, I think it makes the intent clearer.

src/utils/pagedIterator.ts
```diff
@@ -0,0 +1,85 @@
...
+    let issueRequests: [StorageKey, Codec][];
+    while (true) {
+        // a blank StorageKey is 0x in hex
+        if (startKey.toHex() == "0x") {
+            if (!isRemainingListNonempty(polkadotListings)) {
+                return undefined;
+            }
+            issueRequests = await listPaged(polkadotListings, perPage);
+        } else {
+            if (!isRemainingListNonempty(polkadotListings, startKey)) {
```
danhper, line 25: Looking below, it appears that we get an error if there is nothing to return, so I would just add a `try/catch` and be done with it. 

```typescript
let startKey: StorageKey | undefined;
while (true) {
    try {
        const issueRequests = await listPaged(polkadotListings, perPage, startKey);
        if (issueRequests.length === 0) { // not sure if that's needed or if it always be an error in this case
            break;
        }
        startKey = issueRequests[issueRequests.length - 1][0];
        yield issueRequests.map((v) => v[1] as T);
    } catch (e) {
        break;
    }
}
```

Let me know if I'm missing something.

src/utils/pagedIterator.ts
```diff
@@ -0,0 +1,85 @@
...
+
+async function listPaged<T extends PolkadotCodecType>(
+    // eslint-disable-next-line max-len
+    polkadotListings: AugmentedQuery<ApiTypes, (arg: CodecArg) => Observable<T>> & QueryableStorageEntry<ApiTypes>,
+    perPage: number,
+    startKey?: StorageKey
+): Promise<[StorageKey, Codec][]> {
+    let issueRequests: [StorageKey, Codec][] = [];
+
+    if (startKey) {
```
danhper, line 51: This can be simplified quite a bit:

```typescript
const start = startKey ? startKey.toHex() : undefined;
return await polkadotListings.entriesPaged(
                { pageSize: perPage, startKey: start } as PaginationOptions
            ) as [StorageKey, Codec][];
```

src/utils/pagedIterator.ts
```diff
@@ -0,0 +1,85 @@
...
+    // eslint-disable-next-line max-len
+    polkadotListings: AugmentedQuery<ApiTypes, (arg: CodecArg) => Observable<T>> & QueryableStorageEntry<ApiTypes>,
+    startKey?: StorageKey
+): Promise<boolean> {
+    // requesting ENTRIES using the last key as a cut-off point throws a substrate 
+    // error however, requesting KEYS with the last key only returns an empty array
+    // the latter method can be used to check if all entries were "consumed"
+    let keys: StorageKey[] = [];
+    const minimumPageSize = 1;
+    if (startKey) {
```
danhper, line 75: Same as above, condition can be much simpler.

test/unit/apis/issue.test.ts
```diff
@@ -102,6 +102,30 @@ describe.skip("issue", () => {
...
+            for (let i = 0; i < sentRequests; i++) {
+                const amount = api.createType("Balance", i);
+                await issueAPI.request(amount, bobVaultId);
+            }
+
+            const listingsPerPage = 2;
+            const requestsIterator = issueAPI.getPagedIterator(listingsPerPage);
+            let curr = await requestsIterator.next();
+            let requestCount = 0;
+            while (!curr.done) {
```
danhper, line 122: Hopefully, the following should work too, right?

```typescript
for await (const page of issueAPI.getPagedIterator(listingsPerPage)) {
    requestCount += page.length;
    assert.isTrue(page.length <= listingsPerPage);
}
```

danhper (commented):
I think it's all working but a few things could be simplified quite a bit so I would suggest giving it a quick try before merging.

src/utils/pagedIterator.ts
savudani8: That was my approach initially and it is indeed much simpler. I added a comment on line 70 justifying this solution:
```
// requesting ENTRIES using the last key as a cut-off point throws a substrate 
// error however, requesting KEYS with the last key only returns an empty array
// the latter method can be used to check if all entries were "consumed"
```.

So I need to check using the keys first

PR #52, opened 2020-10-07
Check for issue.ExecuteIssue in isExecutionSucessful

savudani8:


src/apis/issue.ts
```diff
@@ -62,12 +62,19 @@ export class DefaultIssueAPI implements IssueAPI {
...
+        // emitted event is system.ExtrinsicFailed
+        if (events.length < 2) {
+            return false;
         }
-        return false;
+
+        let foundExecuteIssueEvent = false;
+        events.forEach(({ event: { method, section } }) => {
+            if (section === "issue" && method === "ExecuteIssue") {
+                foundExecuteIssueEvent = true;
```
nud3l, line 74: replace here with `return true;` to only loop until we find the event

src/apis/issue.ts
```diff
@@ -62,12 +62,19 @@ export class DefaultIssueAPI implements IssueAPI {
...
+            }
+        });
+        return foundExecuteIssueEvent;
```
nud3l, line 77: replace with `return false`

src/apis/issue.ts
```diff
@@ -62,12 +62,19 @@ export class DefaultIssueAPI implements IssueAPI {
         // - issue.ExecuteIssue
         // - system.ExtrinsicSuccess
 
-        // Assuming `system.NewAccount` only occurs on
-        // first execution, at least 5 events are emitted
-        if (events.length >= 5) {
-            return true;
+        // Whenever there is a failure, the only
+        // emitted event is system.ExtrinsicFailed
+        if (events.length < 2) {
```
nud3l, line 67: I think you could skip this check since by default it will return false

src/apis/issue.ts
```diff
@@ -62,12 +62,19 @@ export class DefaultIssueAPI implements IssueAPI {
...
+            return false;
         }
-        return false;
+
+        let foundExecuteIssueEvent = false;
```
nud3l, line 71: don't need this anymore if we return true or false directly

nud3l (changes requested):

nud3l (approved):

PR #54, opened 2020-10-08
PoC: Add tx event logging util

savudani8:
This PR showcases the usage of such a utlity, by wrapping _issue_ requests with the `sendLoggedTx` util to log all events to console. See `issue.ts`, in the `async request` method.

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -4,43 +4,26 @@ import { Callback, ISubmittableResult } from "@polkadot/types/types";
 import { EventRecord, DispatchError } from "@polkadot/types/interfaces/system";
 import { ApiPromise } from "@polkadot/api";
 
-declare type AnyFunctionReturningSubmittableExtrinsicApis =
-    SubmittableExtrinsic<ApiTypes>;
-
-interface EventContainer {
-    events: EventRecord[];
-}
-
 export async function sendLoggedTx(
-    transaction: AnyFunctionReturningSubmittableExtrinsicApis,
+    transaction: SubmittableExtrinsic<"promise">,
     signer: AddressOrPair,
-    api: ApiPromise,
-    eventContainer?: EventContainer
-): Promise<void> {
+    api: ApiPromise
+): Promise<EventRecord[]> {
     // When passing { nonce: -1 } to signAndSend the API will use system.accountNextIndex to determine the nonce
-    const unsubscribe: Callback<ISubmittableResult> = await transaction
-        .signAndSend(
-            signer, { nonce: -1 }, (result) => txCallback(unsubscribe, result, api, eventContainer)
-        ) as Callback<ISubmittableResult>;
-}
+    const { unsubscribe, result } = await new Promise((resolve) => {
+        let unsubscribe: () => void;
+        // When passing { nonce: -1 } to signAndSend the API will use system.accountNextIndex to determine the nonce
+        // signAndSend: Promise<() => void>
+        // signAndSend -> signAndSend resolves (we set unsubscribe) -> callback is called
+        transaction
+            .signAndSend(signer, { nonce: -1 }, (result: ISubmittableResult) => resolve({ unsubscribe, result }))
+            .then((u: () => void) => (unsubscribe = u));
+    });
 
-function txCallback(
-    unsubscribe: Callback<ISubmittableResult>,
-    result: ISubmittableResult,
-    api: ApiPromise,
-    eventContainer?: EventContainer
-) {
-    if (!result.status.isFinalized) {
-        return;
-    }
-    if (result.status.isFinalized) {
-        console.log(`Transaction finalized at blockHash ${result.status.asFinalized}`);
-        unsubscribe(result);
-        printEvents(result.events, api);
-        if (eventContainer) {
-            eventContainer.events = result.events;
-        }
-    }
+    console.log(`Transaction finalized at blockHash ${result.status.asFinalized}`);
+    unsubscribe(result);
+    printEvents(result.events, api);
+    return result.events;
 }
 
 function printEvents(events: EventRecord[], api: ApiPromise) {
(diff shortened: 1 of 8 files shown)


src/utils/logger.ts
```diff
@@ -0,0 +1,68 @@
...
+        ) as Callback<ISubmittableResult>;
+}
+
+function txCallback(
+    unsubscribe: Callback<ISubmittableResult>,
+    result: ISubmittableResult,
+    api: ApiPromise,
+    eventContainer: EventContainer
+) {
+    if (result.status.isFinalized) {
```
danhper, line 33: Nitpick, but I highly recommend early returns for this type of things:

```typescript
if (!result.status.isFinalized) {
    return;
}
// do the actual stuff
```

src/utils/logger.ts
```diff
@@ -0,0 +1,68 @@
...
+        console.log(`Transaction finalized at blockHash ${result.status.asFinalized}`);
+        unsubscribe(result);
+        printEvents(result.events, api);
+        eventContainer.events = result.events;
+    }
+}
+
+function printEvents(events: EventRecord[], api: ApiPromise) {
+    let foundErrorEvent = false;
+    events.forEach(({ event }) => {
```
danhper, line 43: Nitpick: if you want to avoid a bit of nesting, you can probably go with

```typescript
events.flatMap(({ event }) => event.data).forEach( (eventData: any) =>)
```

it will require to add
```json
"lib": [
    "ES2019"
]
```

to the tsconfig.json `compilerOptions` part, which I think is fine.

Same comment as above for early return.

src/utils/logger.ts
```diff
@@ -0,0 +1,68 @@
...
+        event.data.forEach(async (eventData: any) => {
```
danhper, line 44: Does this need to be async?

src/utils/logger.ts
```diff
@@ -0,0 +1,68 @@
+import { AddressOrPair } from "@polkadot/api/submittable/types";
+import { SubmittableExtrinsic, ApiTypes } from "@polkadot/api/types";
+import { Callback, ISubmittableResult } from "@polkadot/types/types";
+import { EventRecord, DispatchError } from "@polkadot/types/interfaces/system";
+import { ApiPromise } from "@polkadot/api";
+
+declare type AnyFunctionReturningSubmittableExtrinsicApis =
+    SubmittableExtrinsic<ApiTypes>;
+
+interface EventContainer {
```
danhper, line 10: Are we planning on adding anything to this interface? Otherwise it does feel like passing `events` directly might be enough.

savudani8: In order to return `events` directly, we'd need to wait for the callback function to be called, which isn't the most JavaScript-y way of doing things. I argue that setting the field of the passed EventContainer object is cleaner

danhper: I think the opposite, waiting for callbacks is a very JavaScript-y way to do things.
With the current implementation, we have no way to tell when `events` is set.
I would rather suggest returning a `Promise<EventRecord[]>`.
The `signAndSend` looks like it was designed for rxjs, not for promises, so the API is really nasty but I think It should be doable with something like this

```typescript
const { unsubscribe, result } = await new Promise((resolve) => {
        let unsubscribe: () => void;
        // When passing { nonce: -1 } to signAndSend the API will use system.accountNextIndex to determine the nonce
        return transaction
            .signAndSend(signer, { nonce: -1 }, (result: ISubmittableResult) => resolve({ unsubscribe, result }))
            .then((u: (() => void)) => (unsubscribe = u));
    });
```

As with your solution, we are relying on the fact that the internal implementation should resolve the promise before calling the callback, otherwise both solutions fail.

danhper:
I looked a little more into this.
Here is my proposition for the `sendLoggedTx`

```typescript
export async function sendLoggedTx(
    transaction: SubmittableExtrinsic<"promise">,
    signer: AddressOrPair,
    api: ApiPromise,
): Promise<EventRecord[]> {
    const { unsubscribe, result } = await new Promise<SignAndSendResult>((resolve) => {
        let unsubscribe: () => void;
        // When passing { nonce: -1 } to signAndSend the API will use system.accountNextIndex to determine the nonce
        return transaction
            .signAndSend(signer, { nonce: -1 }, (result) => resolve({ unsubscribe, result }))
            .then((u) => (unsubscribe = u));
    });
    if (!result.status.isFinalized) {
        return [];
    }
    console.log(`Transaction finalized at blockHash ${result.status.asFinalized}`);
    unsubscribe();
    printEvents(result.events, api);
    return result.events;
}
```

As mentioned above, we rely on the internal behavior of `signAndSend`, which is not ideal, but I don't think we have a choice.
The main advantage is that we can await the events.
Please let me know what you think!

savudani8:
> I looked a little more into this.
> Here is my proposition for the `sendLoggedTx`
> 
> ```ts
> export async function sendLoggedTx(
>     transaction: SubmittableExtrinsic<"promise">,
>     signer: AddressOrPair,
>     api: ApiPromise,
> ): Promise<EventRecord[]> {
>     const { unsubscribe, result } = await new Promise<SignAndSendResult>((resolve) => {
>         let unsubscribe: () => void;
>         // When passing { nonce: -1 } to signAndSend the API will use system.accountNextIndex to determine the nonce
>         return transaction
>             .signAndSend(signer, { nonce: -1 }, (result) => resolve({ unsubscribe, result }))
>             .then((u) => (unsubscribe = u));
>     });
>     if (!result.status.isFinalized) {
>         return [];
>     }
>     console.log(`Transaction finalized at blockHash ${result.status.asFinalized}`);
>     unsubscribe();
>     printEvents(result.events, api);
>     return result.events;
> }
> ```
> 
> As mentioned above, we rely on the internal behavior of `signAndSend`, which is not ideal, but I don't think we have a choice.
> The main advantage is that we can await the events.
> Please let me know what you think!

I've read up a bit on Promises now. I'm unsure why we return from the `await new Promise<SignAndSendResult>`, since the promise is resolved through the callback. Also is the presence of `.then((u) => (unsubscribe = u));` purely to wait for the callback to resolve?

PR #55, opened 2020-10-14
Test cleanup

danhper:
I did quite a bit of cleanup in the tests.
Anything that connects to the parachain has been moved to integration tests.  
We still need to build a decent setup for integration tests where we can have reproducible results though.


nud3l:
Can you list here or somewhere the testdata you would need on the parachain side for the integration tests?

test/helpers.ts
```diff
@@ -0,0 +1,37 @@
+import { DispatchError, EventRecord } from "@polkadot/types/interfaces/system";
+import { ApiPromise } from "@polkadot/api";
+
+export function delay(ms: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+export function printEvents(api: ApiPromise, testType: string, events: EventRecord[]): void {
```
savudani8, line 8: I think you no longer need this function since we merged the tx logging util

savudani8 (approved):

PR #58, opened 2020-10-19
Add event logging to txs, add missing custom types

savudani8:


test/integration/apis/issue.test.ts
```diff
@@ -11,7 +11,7 @@ import { defaultEndpoint } from "../../config";
 
 export type RequestResult = { hash: Hash; vault: Vault };
 
-describe("issue", () => {
+describe.skip("issue", () => {
```
danhper, line 14: Could you add a comment about why it needs to be skipped, please?

test/integration/apis/vaults.test.ts
```diff
@@ -9,7 +9,7 @@ import { PolkaBTC, Vault } from "../../../src/interfaces/default";
 import { assert } from "../../chai";
 import { defaultEndpoint } from "../../config";
 
-describe("vaultsAPI", () => {
+describe.skip("vaultsAPI", () => {
```
danhper, line 12: These tests are passing for me, why are we skipping them?

danhper (changes requested):
There's one issue with the import.

Otherwise, could you please only `skip` the tests if they are not passing, and write why they are not passing. It should make things simpler to fix.

test/integration/apis/issue.test.ts
savudani8: All tests except `should fail if no account is set` fail. This might be because I need to create txs in the parachain, but we don't have a reliable way of doing mocking parachain state (the testdata-generator throws an error at the moment for me)

test/integration/apis/vaults.test.ts
savudani8: They are not passing for me because my parachain has a blank state at the moment. There used to be some unit tests that registered a vault and an exchange rate, but that code has been refactored out. The testdata generator fails for me, as mentioned above

savudani8:
I'll push without `skip`ping, hoping that I'll get the testdata generator to work if it does for you

PR #61, opened 2020-10-23
Add listForUser to issue and redeem

danhper:
I think that this should work but the return type is a bit tedious so it might be worth having a data structure, probably wrapping `Redeem`, including the ID. What do you think?

### DIFF
--- a/src/apis/redeem.ts
+++ b/src/apis/redeem.ts
@@ -83,6 +83,12 @@ export class DefaultRedeemAPI {
         return redeemRequests.map((v) => v[1]);
     }
 
+    async listForUser(account: AccountId): Promise<[H256, Redeem][]> {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const customAPIRPC = this.api.rpc as any;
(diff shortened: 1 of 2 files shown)


src/apis/redeem.ts
```diff
@@ -83,6 +83,12 @@ export class DefaultRedeemAPI {
         return redeemRequests.map((v) => v[1]);
     }
 
+    async listForUser(account: AccountId): Promise<[H256, Redeem][]> {
```
nud3l, line 86: The return type here should be a `RedeemRequest`:https://github.com/interlay/polkabtc-js/blob/dev/src/interfaces/definitions.ts#L140

nud3l (commented):
I suggest we return a kv data structure instead of an array. Suggestion:

{ issueId1: IssueRequest, issueId2: IssueRequest, ...}

Same for the redeem requests.

savudani8:
I think it's fine as it is. But to use types to their full potential, we could return a `DecoratedRedeemRequest: {id: H256, redeem: RedeemRequest}` that makes it clear that `H256` is the id.

nud3l (approved):

PR #63, opened 2020-10-26
Dan/parachain param getters

savudani8:
Creates api calls for each of the constants that are automatically generated in `src/interfaces/augment-api-consts.ts` by running the yarn `build` script. 

All these API calls are grouped in the same file, `ConstantsAPI`. The alternative would have been to add these API calls to their respective APIs if they exist (such as `IssueAPI`, `VaultsAPI`), but for many constants (such as those for `System` and `TransactionPayment`) no API exists in the lib. I considered that adding too many of these files would be messy, but am waiting for feedback on this decision

### DIFF
--- a/src/apis/constants.ts
+++ b/src/apis/constants.ts
@@ -1,0 +1,109 @@
+import { Balance, BalanceOf, BlockNumber, Moment, RuntimeDbWeight, Weight } from "@polkadot/types/interfaces/runtime";
+import { DOT } from "@interlay/polkabtc/interfaces/default";
+import { ApiPromise } from "@polkadot/api";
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/polkabtc",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
nud3l, line 3: Please remove the update on the version here.

nud3l (changes requested):

nud3l:
I think the approach by just having everything in one constant file is good. Please note that we added several more constants in the current dev branch, but we can add these in a later PR.

nud3l (approved):

PR #64, opened 2020-10-26
List staked relayers with id

savudani8:
Adds a new `listIncludingIds` function to the `StakedRelayerAPI`, that returns a list of [id, ActiveStakedRelayer]. The regular `list` function simply returns a list of `ActiveStakedRelayer`s

### DIFF
--- a/src/apis/staked-relayer.ts
+++ b/src/apis/staked-relayer.ts
@@ -8,6 +8,7 @@ import { pagedIterator } from "../utils";
 
 export interface StakedRelayerAPI {
     list(): Promise<ActiveStakedRelayer[]>;
+    map(): Promise<Map<AccountId, ActiveStakedRelayer>>;
     getPagedIterator(perPage: number): AsyncGenerator<ActiveStakedRelayer[]>;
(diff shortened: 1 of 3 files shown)


src/apis/staked-relayer.ts
```diff
@@ -5,9 +5,11 @@ import { ApiPromise } from "@polkadot/api";
 import { VaultsAPI, DefaultVaultsAPI } from "./vaults";
 import BN from "bn.js";
 import { pagedIterator } from "../utils";
+import StorageKey from "@polkadot/types/primitive/StorageKey";
 
 export interface StakedRelayerAPI {
     list(): Promise<ActiveStakedRelayer[]>;
+    listIncludingIds(): Promise<[StorageKey, ActiveStakedRelayer][]>;
```
nud3l, line 12: We would want the AccountId instead of the StorageKey. Also, can you please return that as a `Map` similar to the issue/redeem requests?

nud3l (changes requested):

nud3l (approved):

PR #66, opened 2020-10-27
Vault issue requests

savudani8:


src/apis/vaults.ts
```diff
@@ -18,15 +20,30 @@ export interface VaultsAPI {
...
-    constructor(private api: ApiPromise) { }
+    constructor(private api: ApiPromise) { 
+    }
 
     async list(): Promise<Vault[]> {
         const vaultsMap = await this.api.query.vaultRegistry.vaults.entriesPaged({ pageSize: 1 });
         return vaultsMap.map((v) => v[1]);
     }
 
+    async listIssueRequests(vaultId: AccountId): Promise<IssueRequest[]> {
```
nud3l, line 34: Similar to the issue requests for the user, please return a `Map` with the issue ID as the key.

nud3l (dismissed):

src/apis/vaults.ts
nud3l, line 34: Also, rename this function to `mapForVault`

nud3l (changes requested):

src/apis/vaults.ts
savudani8: The name might be misleading unless the programmer looks at the entire function signature - it's unclear that the map contains issue requests (it might as well contain redeem requests). Maybe a more suggestive name would be `mapIssueRequests`

savudani8: We can leave it like this for now and I'll rename in the PR for vault redeem requests

nud3l (approved):

PR #70, opened 2020-10-28
fix: update request type defs

gregdhill:


src/apis/constants.ts
```diff
@@ -7,9 +7,7 @@ import { WeightToFeeCoefficient } from "@polkadot/types/interfaces/support";
 
 export interface ConstantsAPI {
     getDotExistentialDeposit(): Balance;
-    getIssuePeriod(): BlockNumber;
     getPolkaBtcExistentialDeposit(): Balance;
-    getReplacePeriod(): BlockNumber;
```
savudani8, line 12: Are these constants no longer part of the parachain? I didn't have the most up-to-date version when adding this API, because of the compilation time

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/polkabtc",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
savudani8, line 3: @nud3l suggested we remove this version bump from the commits. In my opinion we should have this updated on the dev chain too, not only in the tag

src/apis/constants.ts
gregdhill: They're now storage fields so they can be actively updated.

package.json
nud3l: Yeah, let's update this in this PR since it will also update the parachain. But please bump to 0.2.3 since the latest release is 0.2.2

nud3l (approved):

PR #72, opened 2020-10-29
feat: update json rpc types && implement vault client

gregdhill:

### DIFF
--- a/src/http/client.ts
+++ b/src/http/client.ts
@@ -1,0 +1,60 @@
+import fetch, { RequestInfo } from "node-fetch";
+
+type RequestParams = Array<string> | undefined;
+type JsonRpcId = number | string;
+
+export interface JsonRpcRequest {
+    jsonrpc: string;
+    method: string;
+    params: RequestParams;
+    id?: JsonRpcId | null;
+}
+
+export interface JsonRpcError {
+    code: number;
+    message: string;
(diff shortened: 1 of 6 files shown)


src/http/client.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export class JsonRpcClient {
+    url: string;
+
+    constructor(url: string) {
+        this.url = url;
+    }
+
+    async post(method: string, params?: RequestParams): Promise<JsonRpcResponse> {
+        const id = Math.random().toString(16).substring(7);
```
savudani8, line 34: This id generation might cause collisions - if collisions are not a problem, why do our RpcRequests use this parameter?

src/http/client.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+interface JsonRpcResponse {
+    jsonrpc: string;
+    result?: string;
+    error?: JsonRpcError;
+    id?: JsonRpcId | null;
+}
+
+export class JsonRpcClient {
```
savudani8, line 26: Nice refactoring 👍

savudani8 (approved):

src/http/client.ts
gregdhill: https://www.jsonrpc.org/specification#request_object

PR #74, opened 2020-10-30
Version 0.2.3

savudani8:


src/apis/constants.ts
```diff
@@ -0,0 +1,98 @@
+import { Balance, BalanceOf, BlockNumber, Moment, RuntimeDbWeight, Weight } from "@polkadot/types/interfaces/runtime";
+import { DOT } from "@interlay/polkabtc/interfaces/default";
+import { ApiPromise } from "@polkadot/api";
+import { u32, u64 } from "@polkadot/types/primitive";
+import { Vec } from "@polkadot/types/codec";
+import { WeightToFeeCoefficient } from "@polkadot/types/interfaces/support";
+
+export interface ConstantsAPI {
```
danhper, line 8: If these are constants, maybe we could make them properties instead of functions and implement them as getters?
Let me know what you think.

nud3l (approved):

src/apis/issue.ts
```diff
@@ -114,6 +114,15 @@ export class DefaultIssueAPI implements IssueAPI {
         return issueRequests.map((v) => v[1]);
     }
 
+    async mapForUser(account: AccountId): Promise<Map<H256, IssueRequest>> {
```
danhper, line 117: I would prefer to explicitly state what the map key is in the function name. `indexedByIdForUser` or something like that?

src/apis/staked-relayer.ts
```diff
@@ -34,6 +35,22 @@ export class DefaultStakedRelayerAPI implements StakedRelayerAPI {
         return activeStakedRelayersMap.map((v) => v[1]);
     }
 
+    async map(): Promise<Map<AccountId, ActiveStakedRelayer>> {
+        const activeStakedRelayers = await this.api.query.stakedRelayers.activeStakedRelayers.entries();
+        const activeStakedRelayerPairs: [
```
danhper, line 40: Is typescript not able to infer types here? Maybe adding `as AccountId` to the `createType` would help?

src/apis/staked-relayer.ts
```diff
@@ -34,6 +35,22 @@ export class DefaultStakedRelayerAPI implements StakedRelayerAPI {
...
+            AccountId,
+            ActiveStakedRelayer
+        ][] = activeStakedRelayers.map((activeStakedRelayer) => [
+            this.api.createType("AccountId", activeStakedRelayer[0]),
+            activeStakedRelayer[1],
+        ]);
+        const activeStakedRelayersMap: Map<AccountId, ActiveStakedRelayer> = new Map<AccountId, ActiveStakedRelayer>();
```
danhper, line 47: Type should be inferred here, no need to make it Java 6 style :smile:

src/apis/vaults.ts
```diff
@@ -18,15 +23,64 @@ export interface VaultsAPI {
...
+     * Fetch the issue requests associated with a vault
+     *
+     * @remarks
+     * Cannot instantiate issueAPI in the constructor, because that would create a dependency loop,
+     * since issueAPI also instantiates the vaultsAPI in its constructor
+     *
+     * @param vaultId - The AccountId of the vault used to filter issue requests
+     * @returns A map with a single key, from the vault AccountId to issue requests involving said vault
+     */
+    async mapIssueRequests(vaultId: AccountId): Promise<Map<AccountId, IssueRequest[]>> {
```
danhper, line 52: I don't think this belongs here. This should be in the issue API, which should allow to filter by vault id.
The fact that we have a circular dependency clearly shows that something is wrong here.

src/apis/vaults.ts
```diff
@@ -70,8 +123,10 @@ export class DefaultVaultsAPI {
     async selectRandomVaultIssue(btc: PolkaBTC): Promise<AccountId> {
         const customAPIRPC = this.api.rpc as any;
         try {
-            const firstVaultWithSufficientCollateral =
-                await customAPIRPC.vaultRegistry.getFirstVaultWithSufficientCollateral(btc);
+            // eslint-disable-next-line max-len
+            const firstVaultWithSufficientCollateral = await customAPIRPC.vaultRegistry.getFirstVaultWithSufficientCollateral(
```
danhper, line 127: I did not catch this before but why do we have a try/catch? If we throw in an `async` function it will reject the promise, so this is redundant I think.

src/http/client.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export class JsonRpcClient {
+    url: string;
+
+    constructor(url: string) {
+        this.url = url;
+    }
+
+    async post(method: string, params?: RequestParams): Promise<JsonRpcResponse> {
+        const id = Math.random().toString(16).substring(7);
```
danhper, line 34: I would make the id an ever incrementing number so that it's half useful for debugging.

src/mock/apis/staked-relayer.ts
```diff
@@ -19,6 +19,15 @@ export class MockStakedRelayerAPI implements StakedRelayerAPI {
         ];
     }
 
+    async map(): Promise<Map<AccountId, ActiveStakedRelayer>> {
+        const registry = new TypeRegistry();
+        const decodedAccountId = "0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d";
+        return new Map([
+            [new GenericAccountId(registry, decodedAccountId), <ActiveStakedRelayer>{ stake: new BN(10.2) as DOT }],
```
danhper, line 26: `BN` does not support float, this will be truncated to 10.
Prefer to pass in strings in the constructor to avoid this type of issues, it would yield a proper error message.

src/utils/pagedIterator.ts
```diff
@@ -41,13 +42,14 @@ async function nextPageLength<T extends PolkadotCodecType>(
...
-    // requesting ENTRIES using the last key as a cut-off point throws a substrate 
+    // requesting ENTRIES using the last key as a cut-off point throws a substrate
     // error however, requesting KEYS with the last key only returns an empty array
     // the latter method can be used to check if all entries were "consumed"
     let keys: StorageKey[] = [];
     const start = startKey ? startKey.toHex() : undefined;
-    keys = await polkadotListings.keysPaged(
-        { pageSize: perPage, startKey: start } as PaginationOptions
-    ) as StorageKey[];
+    keys = (await polkadotListings.keysPaged({
```
danhper, line 50: Why do we need this to be declared beforehand? I think it could be declared here.

test/integration/apis/constants.test.ts
```diff
@@ -0,0 +1,140 @@
+import { ApiPromise } from "@polkadot/api";
+import { assert } from "chai";
+import { ConstantsAPI, DefaultConstantsAPI } from "../../../src/apis/constants";
+import { createPolkadotAPI } from "../../../src/factory";
+import { defaultEndpoint } from "../../config";
+
+describe("Constants", function () {
+    this.timeout(10000); // API can be slightly slow
```
danhper, line 8: Aren't we setting this globally already when running integration tests?

test/integration/apis/constants.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    });
+
+    afterEach(async () => {
+        await api.disconnect();
+    });
+
+    describe("getDotExistentialDeposit", () => {
+        it("should sucessfully return", async () => {
+            const returnValue = await constantAPI.getDotExistentialDeposit();
+            assert.isDefined(returnValue);
```
danhper, line 25: Maybe we can have slightly better tests, for example at least checking the type, and maybe things like greater than 0 or so?

danhper (commented):
I think it looks good overall but there are definitely a few things which would be worth fixing sooner than later.

PR #77, opened 2020-11-02
feat(collateralization): get the total collateralization of the system

nud3l:


savudani8:
Might need to rename `getCollateralization` and `getTotalCollateralization` to `getVaultCollateralization` and `getSystemCollateralization` in the future.

Also, it looks like there's a problem with your commit signature.

savudani8 (approved):

PR #82, opened 2020-11-03
feat: Query btc txs by op_return

savudani8:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/polkabtc",
-  "version": "0.3.2",
+  "version": "0.4.2",
```
nud3l, line 3: Can you update to 0.4.0? Should reset the patch indicator.

src/apis/btc-core.ts
```diff
@@ -1,31 +1,56 @@
-import { BlockApi, Status, TxApi } from "@interlay/esplora-btc-api";
+import { BlockApi, Status, TxApi, ScripthashApi, Transaction } from "@interlay/esplora-btc-api";
 import { AxiosResponse } from "axios";
+import * as bitcoinjs from "bitcoinjs-lib";
 
 const mainnetApiBasePath = "https://blockstream.info/api";
-const testnetApiBasePath = "https://blockstream.info/testnet/api";
+const testnetApiBasePath = "https://electr-testnet.do.polkabtc.io";
```
nud3l, line 6: Why are we declaring this here actually? We should use the definition directly from https://github.com/interlay/esplora-btc-api/blob/master/api.yaml#L8

src/apis/btc-core.ts
```diff
@@ -1,31 +1,56 @@
...
+}
+
 export interface BTCCoreAPI {
     getLatestBlock(): Promise<string>;
     getLatestBlockHeight(): Promise<number>;
     getMerkleProof(txid: string): Promise<string>;
     getTransactionStatus(txid: string): Promise<TxStatus>;
     getTransactionBlockHeight(txid: string): Promise<number | undefined>;
     getRawTransaction(txid: string): Promise<Buffer>;
+    getTxByOpcode(opcode: string): Promise<string>;
```
nud3l, line 41: can you rename to `getTxIdByOpcode` since we don't want to return the full transaction but rather just the id.

nud3l (changes requested):

src/apis/btc-core.ts
savudani8: Good point. Then I'll change the testnet server URL to our node in the esplora wrapper

savudani8: I've had a look over docs and tutorials, and the `servers` object in OpenAPI only serves for showing a drop-down of available servers on swagger. It doesn't seem to be accessible once the API code is generated.
https://idratherbewriting.com/learnapidoc/pubapis_openapi_step3_servers_object.html

Nonetheless, I'll update esplora-btc-api anyways

nud3l: Ah, ok. Thanks, I wasn't aware of that. Then we'll keep it as is!

nud3l (approved):

PR #92, opened 2020-11-07
Vaults - Handle "no tokens issued" error

savudani8:


src/apis/vaults.ts
```diff
@@ -135,12 +142,15 @@ export class DefaultVaultsAPI {
...
      */
-    async getTotalCollateralization(): Promise<number> {
+    async getSystemCollateralization(): Promise<number | undefined> {
         const customAPIRPC = this.api.rpc as any;
         try {
             const collateralization = await customAPIRPC.vaultRegistry.getTotalCollateralization();
             return this.scaleUsingParachainGranularity(collateralization);
         } catch (e) {
+            if (this.isNoTokensIssuedError(e)) {
+                return Promise.resolve(undefined);
```
gregdhill, line 152: I'm not sure about returning undefined here, would we be able to just return `0` instead?

savudani8: We're returning undefined because that's how we handle infinite collateralization in the UI

See here:
https://gitlab.com/interlay/polkabtc-ui/-/blob/dev/src/common/types/vault.types.ts#L15

https://gitlab.com/interlay/polkabtc-ui/-/blob/dev/src/common/components/vault-table/vault-table.tsx#L141

gregdhill (approved):

PR #94, opened 2020-11-09
feat(btc-relay): get stable confirmations param from btc-relay

gregdhill:


savudani8 (approved):
These rpc functions are quite hard to unit test because polkadot-js does not generate proper types for them. Instead, an integration test for this getter would be useful

nud3l (approved):

PR #121, opened 2020-11-11
feat: add boolean checks for active / inactive stake

gregdhill:

### DIFF
--- a/src/apis/staked-relayer.ts
+++ b/src/apis/staked-relayer.ts
@@ -11,6 +11,8 @@ export interface StakedRelayerAPI {
     map(): Promise<Map<AccountId, ActiveStakedRelayer>>;
     getPagedIterator(perPage: number): AsyncGenerator<ActiveStakedRelayer[]>;
     get(activeStakedRelayerId: AccountId): Promise<ActiveStakedRelayer>;
+    isStakedRelayerActive(stakedRelayerId: AccountId): Promise<boolean>;
+    isStakedRelayerInactive(stakedRelayerId: AccountId): Promise<boolean>;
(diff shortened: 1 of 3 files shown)


src/apis/staked-relayer.ts
```diff
@@ -57,6 +59,16 @@ export class DefaultStakedRelayerAPI implements StakedRelayerAPI {
...
     }
 
+    async isStakedRelayerActive(stakedRelayerId: AccountId): Promise<boolean> {
+        const active = await this.api.query.stakedRelayers.activeStakedRelayers(stakedRelayerId);
+        return active.stake.gt(new BN(0));
+    }
+
+    async isStakedRelayerInactive(stakedRelayerId: AccountId): Promise<boolean> {
+        const inactive = await this.api.query.stakedRelayers.inactiveStakedRelayers(stakedRelayerId);
+        return inactive.stake.gt(new BN(0));
```
savudani8, line 69: If the relayer is inactive, shouldn't the stake be less than the `STAKED_RELAYER_STAKE` (https://interlay.gitlab.io/polkabtc-spec/spec/staked-relayers.html#function-sequence)?

So something like `inactive.stake.lt(constants.getStakedRelayersMinimumStake())`

gregdhill: So `STAKED_RELAYER_STAKE` is only used for registration, after which if the storage entry for `ActiveStakedRelayers` or `InactiveStakedRelayers` contains a non-zero stake then the account is registered. An account may set themselves as idle (which moves the staked relayer storage entry to inactive) or more likely they are inactive when 'bonding' - we wait a period of blocks before automatically activating the staked relayer.

savudani8 (approved):

PR #128, opened 2020-11-13
fix(vaults): Apply BalanceWrapper to all currencies

savudani8:


src/apis/vaults.ts
```diff
@@ -224,9 +225,12 @@ export class DefaultVaultsAPI {
     async selectRandomVaultIssue(btc: PolkaBTC): Promise<AccountId> {
         const customAPIRPC = this.api.rpc as any;
         try {
+            const btcWrapped = {
```
gregdhill, line 228: Can you wrap this functionality in a shared function?

gregdhill (approved):

PR #130, opened 2020-11-13
feat(redeem): Subscribe to redeem expiry

savudani8:


src/apis/redeem.ts
```diff
@@ -96,6 +111,23 @@ export class DefaultRedeemAPI {
         return mapForUser;
     }
 
+    async subscribeToRedeemExpiry(callback: (requestRedeemId: string) => any): Promise<void> {
```
gregdhill, line 114:
```suggestion
    async subscribeToRedeemExpiry(callback: (requestRedeemId: string) => void): Promise<void> {
```

src/apis/redeem.ts
```diff
@@ -96,6 +111,23 @@ export class DefaultRedeemAPI {
...
+        const _unsubscribe = await this.api.rpc.chain.subscribeNewHeads(async (header: Header) => {
```
gregdhill, line 115: Perhaps we should return this so the UI can `unsubscribe` on unmount.

gregdhill (changes requested):

src/apis/redeem.ts
```diff
@@ -96,6 +111,24 @@ export class DefaultRedeemAPI {
         return mapForUser;
     }
 
+    async subscribeToRedeemExpiry(callback: (requestRedeemId: string) => void): Promise<() => void> {
+        const unsubscribe = await this.api.rpc.chain.subscribeNewHeads(async (header: Header) => {
```
gregdhill, line 115:
```suggestion
        return this.api.rpc.chain.subscribeNewHeads(async (header: Header) => {
```

gregdhill (approved):

PR #145, opened 2020-11-18
Get issue period

filip88:


src/apis/btc-relay.ts
```diff
@@ -10,6 +11,7 @@ export interface BTCRelayAPI {
     getLatestBlock(): Promise<H256Le>;
     getLatestBlockHeight(): Promise<u32>;
     verifyTransactionInclusion(txid: string, confirmations?: number, insecure?: boolean): Promise<void>;
+    getParachainBlockHeight(): Promise<BlockNumber>;
```
nud3l, line 14: We have this query already in the `system` API: https://github.com/interlay/polkabtc-js/blob/dev/src/apis/system.ts#L11

src/apis/issue.ts
```diff
@@ -129,6 +131,10 @@ export class DefaultIssueAPI implements IssueAPI {
         return pagedIterator<IssueRequest>(this.api.query.issue.issueRequests, perPage);
     }
 
+    async getIssuePeriod(): Promise<BlockNumber> {
+        return (await this.api.query.issue.getIssuePeriod() as BlockNumber);
```
nud3l, line 135: Shouldn't it be `return await this.api.query.issue.issuePeriod();` Best would be to add an integration test to verify that the call is correct here: https://github.com/interlay/polkabtc-js/blob/dev/test/integration/apis/issue.test.ts

nud3l (changes requested):

nud3l (approved):

PR #150, opened 2020-11-20
feat: Add optional amount and recipient params to op_return query

savudani8:
Tested using an op_return that is part of 2 testnet polkaBTC txs, but only the first one has the queried `amount` parameter:
- https://blockstream.info/testnet/tx/f5bcaeb5181154267bf7d05901cc8c2f647414a42126c3aee89e01a2c905ae91?expand
- https://blockstream.info/testnet/tx/4b1900dc48aaa9fa84a340e94aa21d20b54371d19ea6b8edd68a558cd36afdd0?expand

### DIFF
--- a/src/apis/btc-core.ts
+++ b/src/apis/btc-core.ts
@@ -149,8 +149,8 @@ export class DefaultBTCCoreAPI implements BTCCoreAPI {
                 return false;
             }
             if (amountAsBTC) {
-                const amountAsSatoshi = btcToSat(amountAsBTC);
-                if (amountAsSatoshi !== vout.value?.toString()) {
+                const expectedBtcAsSatoshi = Number(btcToSat(amountAsBTC));
+                if (vout.value === undefined || expectedBtcAsSatoshi > vout.value) {
                     return false;
                 }
             }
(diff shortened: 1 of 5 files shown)


src/apis/btc-core.ts
```diff
@@ -125,14 +130,32 @@ export class DefaultBTCCoreAPI implements BTCCoreAPI {
...
 
-        return txs[0].txid;
+    private txOutputHasRecipientAndAmount(vout: VOut, recipientAddress?: string, amountAsBTC?: string): boolean {
+        if (recipientAddress) {
+            if (recipientAddress !== vout.scriptpubkey_address) {
+                return false;
+            }
+            if (amountAsBTC) {
+                const amountAsSatoshi = btcToSat(amountAsBTC);
+                if (amountAsSatoshi !== vout.value?.toString()) {
```
nud3l, line 153: We allow users on the parachain to "overpay" the vault. So instead of checking for equality here, please check if the sent amount is greater than the expected amount.

nud3l (changes requested):

nud3l (approved):

PR #159, opened 2020-11-26
feat(vault): Get secure collateral threshold

savudani8:


nud3l (changes requested):
Let's not expose this in the lib. Instead this should be used to compute the "capacity", i.e. max amount of PolkaBTC that can be issued. 

So something like

`getMaxPolkaBTC(): Promise<string>`

This will need to calculate the totalLockedDOT divided by the exchange rate and dived by the secure threshold (so totalLockedDOT/exchangeRate/secureCollateralThreshold).

src/apis/vaults.ts
```diff
@@ -209,6 +212,20 @@ export class DefaultVaultsAPI {
         return new UInt(new TypeRegistry(), 0) as PolkaBTC;
     }
 
+    private async getSecureCollateralThreshold(): Promise<u128> {
+        return await this.api.query.vaultRegistry.secureCollateralThreshold();
+    }
+
+    async getIssuablePolkaBTC(): Promise<string> {
+        const collateral = new DefaultCollateralAPI(this.api);
+        const totalLockedDot = await collateral.totalLockedDOT();
```
nud3l, line 221: Under the hood, this will return the total locked dot denoted in planck. You need to convert this into dot before doing the calculation by multiplying with DOT_IN_PLANCK constant from utils.

src/apis/vaults.ts
```diff
@@ -209,6 +212,20 @@ export class DefaultVaultsAPI {
...
+        const oracle = new DefaultOracleAPI(this.api);
+        const exchangeRate = await oracle.getExchangeRate();
+        const exchangeRateU128 = this.api.createType("u128", exchangeRate);
```
nud3l, line 224: Why create a type via this call? If you need a `Big` like type, better to use the `Big` constructor like `new Big(exchangeRate)`. The polkadot type convert has practically no error handling, unfortuently.

nud3l (approved):

nud3l:
Minor nit: please also bump the version number.

src/apis/vaults.ts
savudani8: I'm using `createType` because this is the approach recommended by the documentation. `createType` ultimately calls `createTypeUnsafe` function, which handles errors as described in the header comments: https://github.com/polkadot-js/api/blob/master/packages/types/src/create/createType.ts#L62

nud3l (approved):

PR #163, opened 2020-12-02
refactor: use explicit btc address payload type

gregdhill:


package.json
```diff
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "@interlay/esplora-btc-api": "^0.3.3",
-    "@interlay/polkabtc-types": "^0.1.20",
+    "@interlay/polkabtc-types": "0.2.0-rc2",
```
nud3l, line 35: Should be just 0.2.0.

src/http/vault.ts
```diff
@@ -67,13 +69,13 @@ export class VaultClient extends JsonRpcClient {
...
-        const btcAddress = new this.constr["H160"](this.registry, hash);
+    async registerVault(collateral: string): Promise<BtcAddress> {
         const request = new this.constr["RegisterVaultJsonRpcRequest"](this.registry, {
             collateral: new BN(collateral),
-            btc_address: btcAddress,
         });
-        await this.post("register_vault", [request.toHex()]);
+        let response = await this.post("register_vault", [request.toHex()]);
+        const result = new this.constr["RegisterVaultJsonRpcResponse"](this.registry, response.result);
+        return result.address;
```
nud3l, line 78: When interacting with the UI, best is to use strings and not parachain types so we keep the conversion in the lib.

src/http/vault.ts
```diff
@@ -90,12 +92,10 @@ export class VaultClient extends JsonRpcClient {
         await this.post("withdraw_collateral", [request.toHex()]);
     }
 
-    async updateBtcAddress(hash: string): Promise<void> {
-        const btcAddress = new this.constr["H160"](this.registry, hash);
-        const request = new this.constr["UpdateBtcAddressJsonRpcRequest"](this.registry, {
-            address: btcAddress,
-        });
-        await this.post("update_btc_address", [request.toHex()]);
+    async updateBtcAddress(): Promise<BtcAddress> {
```
nud3l, line 95: Also here, let's return the address string instead of the BtcAddress type.

nud3l (approved):

src/apis/issue.ts
```diff
@@ -5,21 +5,37 @@ import { EventRecord } from "@polkadot/types/interfaces/system";
...
+
 export interface IssueAPI {
     request(amount: PolkaBTC, vaultId?: AccountId, griefingCollateral?: DOT): Promise<RequestResult>;
-    execute(issueId: H256, txId: H256Le, txBlockHeight: u32, merkleProof: Bytes, rawTx: Bytes): Promise<boolean>;
+    execute(issueId: H256, txId: H256Le, merkleProof: Bytes, rawTx: Bytes): Promise<boolean>;
     cancel(issueId: H256): Promise<void>;
     setAccount(account?: AddressOrPair): void;
     getGriefingCollateral(): Promise<DOT>;
-    list(): Promise<IssueRequest[]>;
+    list(network: Network): Promise<IssueRequestExt[]>;
```
nud3l, line 35: We already pass the network on constructing the basic API (https://github.com/interlay/polkabtc-js/blob/dev/src/polkabtc-api.ts#L52). Let's store the network as part of the API and get the network from there so we don't have to submit it as a parameter.

nud3l (approved):

src/apis/issue.ts
gregdhill: Ah yes good catch!

nud3l (approved):

PR #165, opened 2020-12-08
Dan/integration tests

savudani8:


src/apis/btc-core.ts
```diff
@@ -165,4 +192,44 @@ export class DefaultBTCCoreAPI implements BTCCoreAPI {
...
+        if (!this.client) {
+            throw new Error("Client needs to be initialized before usage");
+        }
+        const paidOutput = {} as any;
+        paidOutput[receiver] = amount;
+        const raw = await this.client.command("createrawtransaction", [], [{ data: data }, paidOutput]);
+        const funded = await this.client.command("fundrawtransaction", raw);
+        const signed = await this.client.command("signrawtransactionwithwallet", funded.hex);
+        const response = await this.broadcastRawTransaction(signed.hex);
+        await delay(1000);
```
nud3l, line 213: What are these delays used for? I guess on regtest you would need to mine a block right after to have the tx included. If there is good reason for this, please add a comment

src/apis/btc-core.ts
```diff
@@ -165,4 +192,44 @@ export class DefaultBTCCoreAPI implements BTCCoreAPI {
...
+        return {
+            txid: txid,
+            rawTx: signed.hex,
+        };
+    }
+
+    async mineBlocks(n: number) {
+        const newWalletAddress = await this.client.command("getnewaddress");
+        const minedTxs = await this.client.command("generatetoaddress", n, newWalletAddress);
+        const relayPeriodWithBuffer = 6500;
```
nud3l, line 224: also would be nice to know what this relay period is with a line comment

test/integration/apis/btc-core.test.ts
```diff
@@ -76,27 +76,6 @@ describe("BTCCore testnet", function () {
             assert.deepEqual(raw_tx, raw);
         });
     });
-
-    describe("getTxByOpcode", () => {
```
nud3l, line 80: for these integration tests, could you update the btc-core API to point to testnet? I assume those are deleted because the tests are not passing with using the local electrs instance in the docker container.

test/integration/apis/btc-core.test.ts
```diff
@@ -113,13 +92,4 @@ describe("BTCCore regtest", function () {
     afterEach(async () => {
         await api.disconnect();
     });
-
-    describe("getTxByOpcode", () => {
-        it("should return correct tx id", async () => {
-            // FIXME: generate bitcoin transaction with OP_RETURN value
```
nud3l, line 119: could you create a OP_RETURN tx here that you then check if it's included? I'd like to have that integration test because we had some major issues on the MS2 review that the transaction status caused issues.

test/integration/apis/issue.test.ts
```diff
@@ -51,32 +62,13 @@ describe("issue", () => {
...
             keyring = new Keyring({ type: "sr25519" });
-            bob = keyring.addFromUri("//Bob");
             alice = keyring.addFromUri("//Alice");
             issueAPI.setAccount(alice);
-            const bobVaultId = api.createType("AccountId", bob.address);
-            const amount = api.createType("Balance", 1);
-            const requestResult = await issueAPI.request(amount, bobVaultId);
+            const amount = api.createType("Balance", 100000);
+            const requestResult = await issueAPI.request(amount);
+            console.log("finalized request issue");
```
nud3l, line 71: Remove the console.log

test/integration/apis/issue.test.ts
```diff
@@ -51,32 +62,13 @@ describe("issue", () => {
...
             assert.isTrue(requestResult.hash.length > 0);
```
nud3l, line 72: Better would be to check the request issue event for the actual input data. Maybe best would be to expose the helper functions inside the issue api to check for matching events here?

test/integration/apis/issue.test.ts
```diff
@@ -87,44 +79,101 @@ describe("issue", () => {
...
         });
 
         it("should request if account is set", async () => {
             issueAPI.setAccount(alice);
             const amount = api.createType("Balance", 1);
             const bobVaultId = api.createType("AccountId", bob.address);
-            const requestResult = await issueAPI.request(amount, bobVaultId);
+            const requestResult = await issueAPI.request(amount);
             txHash = requestResult.hash;
             assert.isDefined(txHash);
```
nud3l, line 93: Better would be to check the request issue event for the actual input data. Maybe best would be to expose the helper functions inside the issue api to check for matching events here?

test/integration/apis/redeem.test.ts
```diff
@@ -56,81 +64,84 @@ describe.skip("redeem", () => {
...
+            // redeem
             redeemAPI.setAccount(alice);
-            const amount = api.createType("PolkaBTC", 10);
-            requestResult = await redeemAPI.request(amount, randomDecodedAccountId);
+            const redeemAmountAsBtcString = "0.09";
+            const redeemAmountAsSatoshiString = btcToSat(redeemAmountAsBtcString);
+            const redeemAmountAsSatoshi = api.createType("Balance", redeemAmountAsSatoshiString);
+            const btcAddress = "bcrt1qujs29q4gkyn2uj6y570xl460p4y43ruayxu8ry";
+            const daveVaultId = api.createType("AccountId", dave.address);
+            await redeemAPI.request(redeemAmountAsSatoshi, btcAddress, daveVaultId);
```
nud3l, line 100: At the end of the test, should check if a suitable request redeem event has been fired.

test/integration/apis/redeem.test.ts
```diff
@@ -56,81 +64,84 @@ describe.skip("redeem", () => {
...
-            let requestCount = 0;
-            while (!curr.done) {
-                requestCount += curr.value.length;
-                assert.isTrue(curr.value.length <= listingsPerPage);
-                curr = await requestsIterator.next();
-            }
-            assert.equal(requestCount, sentRequests);
-        });
-    });
+            // request issue
```
nud3l, line 75: Since you are reusing the same code below you can potentially move this to a separate function to avoid duplication.

test/integration/apis/issue.test.ts
```diff
@@ -87,44 +79,101 @@ describe("issue", () => {
...
+            const amountAsBtcString = "0.001";
+            const amountAsSatoshiString = btcToSat(amountAsBtcString);
+            if (amountAsSatoshiString === undefined) {
+                fail();
+            }
+            const amountAsSatoshi = api.createType("Balance", amountAsSatoshiString);
+            const requestResult = await issueAPI.request(amountAsSatoshi);
+            assert.isTrue(requestResult.hash.length > 0);
+
+            // send btc tx
```
nud3l, line 132: I would move all of this into a util function that you can share among issue and redeem. Something like
```js
send_data_btc_tx_and_mine(string: data, string: amount, string: recipient)
```

test/integration/apis/issue.test.ts
```diff
@@ -87,44 +79,101 @@ describe("issue", () => {
...
+            if (vaultBtcAddress === undefined) {
+                throw new Error("Undefined vault address returned from RequestIssue");
+            }
+            const txData = await btcCore.broadcastOpReturnTx(vaultBtcAddress, amountAsBtcString, data);
+            await btcCore.mineBlocks(blocksToMine);
+
+            // check issuing worked
+            let finalBalance = await treasuryAPI.balancePolkaBTC(api.createType("AccountId", alice.address));
+            assert.isTrue(finalBalance.toBn().sub(initialBalance.toBn()).eq(new BN(amountAsSatoshiString)));
+        });
```
nud3l, line 144: We need to test the execute isue function here as well. For this set the `--no-issue-execution` flag on the vault client in the docker container.

nud3l (changes requested):

test/integration/apis/issue.test.ts
savudani8: I'll add a check in the issue API that throws an error if events are not ok, so this will automatically invalidate any failing tests

test/integration/apis/redeem.test.ts
savudani8: I'll add a check in the redeem API, same as for issue

nud3l (approved):

PR #166, opened 2020-12-10
feat: Partial fee model implementation

savudani8:


src/apis/issue.ts
```diff
@@ -37,6 +37,7 @@ export interface IssueAPI {
     mapForUser(account: AccountId): Promise<Map<H256, IssueRequestExt>>;
     getRequestById(issueId: string | Uint8Array | H256): Promise<IssueRequestExt>;
     getIssuePeriod(): Promise<BlockNumber>;
+    getFeesToPay(account: AccountId): Promise<DOT>;
```
nud3l, line 40: I'd suggest we remove or make the `account` optional and treat it like on the other api calls where we infer the account from the signer.
Second, we need to add the input amount (PolkaBTC) as the fees depend on this.
Third, the fees are paid in PolkaBTC, so this needs to return PolkaBTC.

src/apis/redeem.ts
```diff
@@ -39,6 +38,7 @@ export interface RedeemAPI {
     getRequestById(redeemId: string | Uint8Array | H256): Promise<RedeemRequestExt>;
     subscribeToRedeemExpiry(account: AccountId, callback: (requestRedeemId: string) => void): Promise<() => void>;
     getDustValue(): Promise<PolkaBTC>;
+    getFeesToPay(account: AccountId): Promise<DOT>;
```
nud3l, line 41: Same points as on issue.

src/apis/staked-relayer.ts
```diff
@@ -24,6 +24,12 @@ export interface StakedRelayerAPI {
     getAllActiveStatusUpdates(): Promise<Array<{ id: u256; statusUpdate: StatusUpdate }>>;
     getAllInactiveStatusUpdates(): Promise<Array<{ id: u256; statusUpdate: StatusUpdate }>>;
     getAllStatusUpdates(): Promise<Array<{ id: u256; statusUpdate: StatusUpdate }>>;
+    getFees(stakedRelayerId: AccountId): Promise<PolkaBTC>;
+    getSLA(stakedRelayerId: AccountId): Promise<number>;
```
nud3l, line 28: SLA and score is the same thing in the current model. So we only need one of the two. All the scoring additions will be handled in the parachain since we need to have them onchain for fee payouts.

src/apis/staked-relayer.ts
```diff
@@ -24,6 +24,12 @@ export interface StakedRelayerAPI {
...
+    getScore(stakedRelayerId: AccountId): Promise<string>;
+    getTotalScore(): Promise<string>;
```
nud3l, line 30: Not sure what the total score would be? The score among all relayers? If yes, then we don't really need this.

src/apis/staked-relayer.ts
```diff
@@ -24,6 +24,12 @@ export interface StakedRelayerAPI {
...
+    getTotalReward(): Promise<string>;
```
nud3l, line 31: What is this?

src/apis/staked-relayer.ts
```diff
@@ -24,6 +24,12 @@ export interface StakedRelayerAPI {
...
+    getReward(stakedRelayerId: AccountId): Promise<string>;
```
nud3l, line 32: Not sure how this is different from the getFees? IMO probably don't need that.

src/apis/staked-relayer.ts
```diff
@@ -24,6 +24,12 @@ export interface StakedRelayerAPI {
     getAllActiveStatusUpdates(): Promise<Array<{ id: u256; statusUpdate: StatusUpdate }>>;
     getAllInactiveStatusUpdates(): Promise<Array<{ id: u256; statusUpdate: StatusUpdate }>>;
     getAllStatusUpdates(): Promise<Array<{ id: u256; statusUpdate: StatusUpdate }>>;
```
nud3l, line 26: we need one additional call which is the max SLA that is going to tell you your relative score. So like you acheived 90/100.

src/apis/vaults.ts
```diff
@@ -279,6 +286,53 @@ export class DefaultVaultsAPI {
...
+
+        return feesInUSD.div(lockedUsdValue).toString();
+    }
+
+    async getSLA(_vaultId: AccountId): Promise<number> {
+        // TODO: get real value from backend
+        return 62;
+    }
+
+    async getSlashableCollateral(vaultId: AccountId): Promise<string> {
```
nud3l, line 319: This logic will move to the parachain as well, so a simple getter here is sufficient. But we'll need the input PolkaBTC amount.

nud3l (changes requested):

src/apis/staked-relayer.ts
savudani8: I followed the fee model document Screenshot from 2020-12-10 17-01-50

src/apis/staked-relayer.ts
savudani8: See the screenshot attached

src/apis/staked-relayer.ts
savudani8: See the screenshot attached

src/apis/staked-relayer.ts
savudani8: See the screenshot attached

src/apis/vaults.ts
savudani8: Don't we already know the staked relayer's collateral? I'm not sure why we need a PolkaBTC input

nud3l: This is to calculate how much the vault can be slashed for a certain redeem request. Say Alice tries to redeem 1 PolkaBTC. It will depend on the vault's SLA how much collateral will be slashed for that specific request.

src/apis/staked-relayer.ts
nud3l: Yes, but that is all hidden in the UI. My bad on saying they are the same. However, the score and reward function will not be exposed externally. This is only for the internal computations in the parachain. In the UI and lib, all we need is the fees earned by the staked relayer and the SLA score.

src/apis/staked-relayer.ts
```diff
@@ -24,6 +24,8 @@ export interface StakedRelayerAPI {
...
+    getFees(stakedRelayerId: AccountId): Promise<PolkaBTC>;
+    getSLA(stakedRelayerId: AccountId): Promise<number>;
 }
```
nud3l, line 29: need to add the getMaxSLA to the api interface as well.

src/apis/vaults.ts
```diff
@@ -34,14 +34,22 @@ export interface VaultsAPI {
     selectRandomVaultIssue(btc: PolkaBTC): Promise<AccountId>;
     selectRandomVaultRedeem(btc: PolkaBTC): Promise<AccountId>;
     getIssuablePolkaBTC(): Promise<string>;
+    getLiquidationCollateralThreshold(): Promise<u128>;
+    getPremiumRedeemThreshold(): Promise<u128>;
+    getFees(vaultId: AccountId): Promise<PolkaBTC>;
+    getAPY(vaultId: AccountId): Promise<string>;
+    getSLA(vaultId: AccountId): Promise<number>;
+    getSlashableCollateral(vaultId: AccountId, amount: PolkaBTC): Promise<string>;
 }
```
nud3l, line 43: please also add the getMaxSLA to the interface

nud3l (changes requested):

nud3l (approved):

PR #168, opened 2020-12-14
fee model: update version of polkabtc types, added cancelled and fee …

filip88:
…fields on requests


nud3l:
please also make sure to do the correct format for the commit messages. It would be feat(fee model): update version of polkabtc types, added cancelled and fee

nud3l (approved):

PR #171, opened 2020-12-14
fee calls

savudani8:


src/apis/oracle.ts
```diff
@@ -1,15 +1,14 @@
...
 import { Moment } from "@polkadot/types/interfaces/runtime";
 import { u128 } from "@polkadot/types/primitive";
-import { BTC_IN_SAT, DOT_IN_PLANCK, sendLoggedTx } from "../utils";
+import { sendLoggedTx } from "../utils";
 import Big from "big.js";
 import { AddressOrPair } from "@polkadot/api/types";
 
 const defaultFeedName = "DOT/BTC";
-const granularity = 5;
+const granularity = 2;
```
nud3l, line 11: Granularity in the parachain is 5. Why 2?

EDIT: Ok, I see you tried to simplify the conversion. I would keep as is. It is possible that the currencies we are dealing with would change their denomation. Unlikely, but it already happened once for Polkadot so it makes it somewhat cleaner to have it that way. We could move the granularity to a constant into the `constants` file though.

src/apis/vaults.ts
```diff
@@ -296,40 +298,44 @@ export class DefaultVaultsAPI {
...
-        const feesInUSD = fees;
-
-        const lockedDot = await this.collateralAPI.balanceLockedDOT(vaultId);
-        // TODO convert from DOT to USD
-        const lockedUsdValue = lockedDot;
-
-        return feesInUSD.div(lockedUsdValue).toString();
+        const feesBig = new Big(fees.toString());
+        const dotToBtcRate = await this.oracleAPI.getExchangeRate();
+        const feesInDot = feesBig.div(new Big(dotToBtcRate));
```
nud3l, line 312: You need to multiply here. If the fees were e.g. 1 PolkaBTC and the exchange rate is e.g. 3855.55 DOT/BTC you'd ned to multiply to get the DOT amount.

nud3l (changes requested):

src/apis/oracle.ts
savudani8: The previous granularity was broken and the Oracle test was being skipped. If you recall, the getExchangeRate function returns 100 or 1000 after the rate is set to 1. The current implementation might be an oversimplification, but the previous one was wrong

src/apis/vaults.ts
savudani8: I'll update, but if we're using DOT/BTC should the rate be something small, like `0.00026455` (https://uk.tradingview.com/symbols/DOTBTC/)?

src/apis/oracle.ts
nud3l: Right, but then let's fix the right things. The granularity on the parachain is 5. Somehow we should reach the correct exchange rate without having inconsistencies between the granularities on the lib and the chain.

src/apis/vaults.ts
nud3l: sorry, the other way round. We have BTC to DOT which is set to something like 3855.55. And then you need to multiply to get from BTC to DOT.

savudani8: Since it is the other way around, we should change the feed name returned by the client: https://github.com/interlay/polkabtc-js/blob/dev/test/unit/apis/oracle.test.ts#L58

nud3l (approved):

PR #188, opened 2020-12-18
feat(fee-model): split the rewards into PolkaBTC and DOT

nud3l:


src/apis/staked-relayer.ts
```diff
@@ -156,21 +157,36 @@ export class DefaultStakedRelayerAPI implements StakedRelayerAPI {
...
     async getAPY(stakedRelayerId: string): Promise<string> {
-        const fees = await this.getFees(stakedRelayerId);
-        const feesBig = new Big(fees.toString());
-        const dotToBtcRate = await this.oracleAPI.getExchangeRate();
-        const feesInDot = feesBig.mul(new Big(dotToBtcRate));
-        const parsedStakedRelayerId = this.api.createType("AccountId", stakedRelayerId);
-        const lockedDot = await this.collateralAPI.balanceLockedDOT(parsedStakedRelayerId);
-        const lockedDotBig = new Big(lockedDot.toString());
-        return feesInDot.div(lockedDotBig).toString();
+        const parsedVaultId = this.api.createType("AccountId", stakedRelayerId);
```
savudani8, line 182: This should be renamed to `parsedStakedRelayerId`

src/apis/staked-relayer.ts
```diff
@@ -156,21 +157,36 @@ export class DefaultStakedRelayerAPI implements StakedRelayerAPI {
...
+        const [feesPolkaBTC, feesDOT, dotToBtcRate, lockedDOT] = await Promise.all([
+            await this.getFeesPolkaBTC(stakedRelayerId),
+            await this.getFeesDOT(stakedRelayerId),
+            await this.oracleAPI.getExchangeRate(),
+            await (await this.collateralAPI.balanceLockedDOT(parsedVaultId)).toString(),
+        ]);
+        return calculateAPY(feesPolkaBTC, feesDOT, lockedDOT, dotToBtcRate);
```
savudani8, line 189: Since we're refactoring the calculation into a common component, it would be great to fetch the `dotToBtcRate` in there as well, such that the new signature becomes `export function calculateAPY(feesPolkaBTC: string, feesDOT: string, lockedDOT: string): string`. We'd only fetch the data that depends on the `stakedRelayerId` here

src/apis/vaults.ts
```diff
@@ -304,20 +306,34 @@ export class DefaultVaultsAPI {
...
-        const dotToBtcRate = await this.oracleAPI.getExchangeRate();
-        const feesInDot = feesBig.mul(new Big(dotToBtcRate));
         const parsedVaultId = this.api.createType("AccountId", vaultId);
-        const lockedDot = await this.collateralAPI.balanceLockedDOT(parsedVaultId);
-        const lockedDotBig = new Big(lockedDot.toString());
-        return feesInDot.div(lockedDotBig).toString();
+        const [feesPolkaBTC, feesDOT, dotToBtcRate, lockedDOT] = await Promise.all([
+            await this.getFeesPolkaBTC(vaultId),
+            await this.getFeesDOT(vaultId),
+            await this.oracleAPI.getExchangeRate(),
```
savudani8, line 333: We can move this into `calculateAPY`, as suggested above

savudani8 (approved):

PR #191, opened 2020-12-18
feat: expose auction and secure collateral threshold getters

gregdhill:

### DIFF
--- a/src/apis/vaults.ts
+++ b/src/apis/vaults.ts
@@ -29,20 +29,18 @@ export interface VaultsAPI {
     mapReplaceRequests(vaultId: AccountId): Promise<Map<H256, ReplaceRequestExt>>;
     getPagedIterator(perPage: number): AsyncGenerator<Vault[]>;
     get(vaultId: AccountId): Promise<Vault>;
-    getVaultCollateralization(
-        vaultId: AccountId,
-        newCollateral?: DOT,
-        onlyIssued?: boolean
-    ): Promise<number | undefined>;
-    getSystemCollateralization(): Promise<number | undefined>;
+    getVaultCollateralization(vaultId: AccountId, newCollateral?: DOT, onlyIssued?: boolean): Promise<Big | undefined>;
+    getSystemCollateralization(): Promise<Big | undefined>;
     getRequiredCollateralForVault(vaultId: AccountId): Promise<DOT>;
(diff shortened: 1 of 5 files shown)


src/apis/vaults.ts
```diff
@@ -40,6 +40,8 @@ export interface VaultsAPI {
     getTotalIssuedPolkaBTCAmount(): Promise<PolkaBTC>;
     selectRandomVaultIssue(btc: PolkaBTC): Promise<AccountId>;
     selectRandomVaultRedeem(btc: PolkaBTC): Promise<AccountId>;
+    getAuctionCollateralThreshold(): Promise<Big>;
```
nud3l, line 43: can you move them down to line 46 so they are with the other thresholds please?

src/apis/vaults.ts
```diff
@@ -250,9 +252,14 @@ export class DefaultVaultsAPI {
         return new UInt(new TypeRegistry(), 0) as PolkaBTC;
     }
 
-    private async getSecureCollateralThreshold(): Promise<Big> {
-        const secureCollateralThresholdU128 = await this.api.query.vaultRegistry.secureCollateralThreshold();
-        return new Big(secureCollateralThresholdU128.toString());
+    async getAuctionCollateralThreshold(): Promise<Big> {
```
nud3l, line 255: can you also move them down to line 308 so we have it with the rest of the thresholds

also, can you transform the other thresholds to return a `Big` please?

src/mock/apis/vaults.ts
```diff
@@ -135,6 +136,14 @@ export class MockVaultsAPI implements VaultsAPI {
         return new UInt(registry, 100) as DOT;
     }
 
+    async getAuctionCollateralThreshold(): Promise<Big> {
```
nud3l, line 139: also, please move to the other thresholds

nud3l (approved):

src/apis/vaults.ts
```diff
@@ -178,10 +180,10 @@ export class DefaultVaultsAPI {
         try {
             const collateralization = newCollateral
                 ? await customAPIRPC.vaultRegistry.getCollateralizationFromVaultAndCollateral(
-                    vaultId,
-                    this.wrapCurrency(newCollateral),
-                    onlyIssued
-                )
+                      vaultId,
```
savudani8, line 183: this indent might cause the lint to fail

PR #192, opened 2020-12-18
Premium redeem fee

filip88:


src/apis/vaults.ts
```diff
@@ -255,6 +256,11 @@ export class DefaultVaultsAPI {
         return new Big(secureCollateralThresholdU128.toString());
     }
 
+    async getPremiumRedeemFee(): Promise<string> {
+        const premiumRedeemFee = await this.api.query.fee.getPremiumRedeemFee();
+        return premiumRedeemFee.toString();
```
nud3l, line 261: You need to scale this with this function before returning: https://github.com/interlay/polkabtc-js/blob/dev/src/utils/encoding.ts#L54

nud3l (changes requested):

PR #196, opened 2020-12-18
fix: custom rpcs return bn not big int

gregdhill:

### DIFF
--- a/src/apis/vaults.ts
+++ b/src/apis/vaults.ts
@@ -181,7 +181,7 @@ export class DefaultVaultsAPI {
                     onlyIssued
                 )
                 : await customAPIRPC.vaultRegistry.getCollateralizationFromVault(vaultId, onlyIssued);
-            return this.scaleUsingParachainGranularity(collateralization);
+            return this.scaleUsingParachainGranularity(new Big(collateralization));
         } catch (e) {
             if (this.isNoTokensIssuedError(e)) {
                 return Promise.resolve(undefined);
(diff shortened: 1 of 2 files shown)


src/apis/vaults.ts
```diff
@@ -176,12 +176,12 @@ export class DefaultVaultsAPI {
         try {
             const collateralization = newCollateral
                 ? await customAPIRPC.vaultRegistry.getCollateralizationFromVaultAndCollateral(
-                    vaultId,
-                    this.wrapCurrency(newCollateral),
-                    onlyIssued
-                )
+                      vaultId,
```
savudani8, line 179: watch out for the indent

src/apis/vaults.ts
```diff
@@ -181,7 +181,7 @@ export class DefaultVaultsAPI {
                     onlyIssued
                 )
                 : await customAPIRPC.vaultRegistry.getCollateralizationFromVault(vaultId, onlyIssued);
-            return this.scaleUsingParachainGranularity(collateralization);
+            return this.scaleUsingParachainGranularity(new Big(collateralization));
```
savudani8, line 184: is Big compatible with the u128 parameter of `scaleUsingParachainGranularity`?

savudani8: We could remove the `private` access modifier and write a unit test

gregdhill: `scaleUsingParachainGranularity` takes `Big` now

gregdhill: yep, let's remember to do that next week :)

savudani8 (approved):

PR #209, opened 2020-12-23
feat: add tx status integration test

nud3l:


test/integration/apis/btc-core.test.ts
```diff
@@ -105,26 +105,53 @@ describe("BTCCore regtest", function () {
 
     let api: ApiPromise;
     let btcCore: BTCCoreAPI;
+    let bitcoinCoreClient: BitcoinCoreClient;
 
     beforeEach(async () => {
         api = await createPolkadotAPI(defaultEndpoint);
         btcCore = new DefaultBTCCoreAPI("http://0.0.0.0:3002");
+        bitcoinCoreClient = new BitcoinCoreClient(
```
savudani8, line 113: I would include this initialization in a  `before` block rather than the `beforeEach`, since `bitcoinCoreClient` holds no state besides the connection

test/integration/apis/btc-core.test.ts
```diff
@@ -169,4 +151,30 @@ describe("BTCCore regtest", function () {
             assert.strictEqual(txid, txData.txid);
         });
     });
+
+    describe("getTxStatus", () => {
+        it("should return 0 confirmations", async () => {
+            const opReturnValue = "01234567891154267bf7d05901cc8c2f647414a42126c3aee89e01a2c905ae91";
```
savudani8, line 157: If we're testing the tx a user sends as part of executeIssue, we would no longer need this opreturn. I'm happy to merge however, since it doesn't influence the logic

savudani8 (approved):

PR #214, opened 2021-01-14
fix: Remove griefing collateral from request replace

savudani8:


package.json
```diff
@@ -33,7 +33,7 @@
   "engineStrict" : true,
   "dependencies": {
     "@interlay/esplora-btc-api": "^0.3.3",
-    "@interlay/polkabtc-types": "0.2.9",
+    "@interlay/polkabtc-types": "0.2.12",
```
nud3l, line 36: This should still be the 0.2.9 version I assume. 0.2.12 has the latest FixedPoint additions in there that we don't want on master yet.

test/integration/apis/issue.test.ts
```diff
@@ -78,21 +78,6 @@ describe("issue", () => {
         });
     });
 
-    describe("load requests", () => {
```
nud3l, line 81: This tests should still work. I guess it is breaking because of the types update.

nud3l (changes requested):

test/integration/apis/issue.test.ts
savudani8: This test was duplicated for some reason

nud3l (approved):

PR #216, opened 2021-01-15
feat: Add faucet interface

savudani8:


src/http/faucet.ts
```diff
@@ -0,0 +1,34 @@
+import {
+    FundAccountJsonRpcRequest,
+} from "../interfaces/default";
+import { getAPITypes } from "../factory";
+import { TypeRegistry,  } from "@polkadot/types";
```
nud3l, line 5: minor nit, no ", " after `TypeRegistry`

test/integration/http/faucet.test.ts
```diff
@@ -0,0 +1,50 @@
+import { ApiPromise, Keyring } from "@polkadot/api";
+import { FaucetClient } from "../../../src/http";
+import { createPolkadotAPI } from "../../../src/factory";
+import { defaultEndpoint } from "../../config";
+import { KeyringPair } from "@polkadot/keyring/types";
+import { AccountData } from "@polkadot/types/interfaces/balances";
+import { assert } from "../../chai";
+import Big from "big.js";
+
+describe.skip("Faucet", function () {
```
nud3l, line 10: why deactivate the test?

nud3l:
In this PR, can you please switch back to the `dev` branches in the docker-compose file? Thanks!

test/integration/http/faucet.test.ts
savudani8: It's "not yet activated" rather than "deactivated". We need the dockerized faucet service in docker-compose first in order to run this test

nud3l: ah, of course. My bad

docker-compose.yml
```diff
@@ -117,7 +117,7 @@ services:
                 testdata-gen --polka-btc-url 'ws://polkabtc:9944' --keyring=alice set-redeem-period --period=50
 
                 # the exchange rate value is 3855.23187 in FixedPoint format
-                testdata-gen --polka-btc-url 'ws://polkabtc:9944' --keyring=bob set-exchange-rate --exchange-rate=3855231870000000000000
+                testdata-gen --polka-btc-url 'ws://polkabtc:9944' --keyring=bob set-exchange-rate --exchange-rate=385523187
```
nud3l, line 120: Which version is correct, with the many 0s or without them?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/polkabtc",
-  "version": "0.6.19",
+  "version": "0.6.20",
```
nud3l, line 3: Since we have breaking API changes, let's bump to 0.7.0

src/apis/btc-relay.ts
```diff
@@ -43,16 +43,14 @@ export class DefaultBTCRelayAPI implements BTCRelayAPI {
...
     async verifyTransactionInclusion(
         txid: string,
         confirmations: number = DEFAULT_STABLE_CONFIRMATIONS,
-        insecure: boolean = false
     ): Promise<void> {
         const merkleProof = await this.btcCore.getMerkleProof(txid);
+        const confirmationsU32 = this.api.createType("u32", confirmations);
         // TODO: change this to RPC call
-        this.api.tx.btcRelay.verifyTransactionInclusion(txid, merkleProof, confirmations, insecure);
+        this.api.tx.btcRelay.verifyTransactionInclusion(txid, merkleProof, confirmationsU32);
```
nud3l, line 54: We didn't need this `u32` conversion before. The only thing that changed was that the underlying type is now an `Option`. Why do we need this type conversion?

docker-compose.yml
savudani8: It used to work with many zeroes because of FixedPoint. This new parachain version broke many tests because vaults no longer had enough collateral. So I assume somethungs changed in how exchange rates are set

src/apis/btc-relay.ts
savudani8: I noticed that, but the compiler was reporting this as an error

nud3l (approved):

PR #219, opened 2021-01-19
feat: Replace op_return queries

savudani8:


src/apis/btc-core.ts
```diff
@@ -39,13 +38,13 @@ export interface BTCCoreAPI {
     getTransactionStatus(txid: string): Promise<TxStatus>;
     getTransactionBlockHeight(txid: string): Promise<number | undefined>;
     getRawTransaction(txid: string): Promise<Buffer>;
-    getTxIdByOpReturn(opReturn: string, recipientAddress?: string, amountAsBTC?: string): Promise<string>;
```
nud3l, line 42: Please leave the OP_RETURN in there. We still need this on redeem.

src/apis/btc-core.ts
```diff
@@ -86,6 +85,49 @@ export class DefaultBTCCoreAPI implements BTCCoreAPI {
         return this.getData(this.txApi.getTxMerkleBlockProof(txid));
     }
 
+    /**
```
nud3l, line 88: NIcely added with doc comments right away, great job!

src/http/vault.ts
```diff
@@ -68,15 +64,6 @@ export class VaultClient extends JsonRpcClient {
         await this.post("request_replace", [request.toHex()]);
     }
 
-    async registerVault(collateral: string, network: bitcoin.Network): Promise<string> {
```
nud3l, line 71: We would still want to give the option to register a vault via the UI. So we would need this one in there.

src/http/vault.ts
```diff
@@ -91,12 +78,6 @@ export class VaultClient extends JsonRpcClient {
         await this.post("withdraw_collateral", [request.toHex()]);
     }
 
-    async updateBtcAddress(network: bitcoin.Network): Promise<string> {
```
nud3l, line 94: Also this one should stay for now.

test/integration/apis/oracle.test.ts
```diff
@@ -27,16 +27,19 @@ describe("OracleAPI", () => {
...
         it("should return oracle info", async () => {
             const info = await oracle.getInfo();
             assert.equal(info.names[0], "Bob");
             assert.isTrue(info.online);
             assert.equal(info.feed, "DOT/BTC");
         });
-    });
 
-    describe("setExchangeRate", () => {
+        it("should have a rate of 3855.23187", async () => {
```
nud3l, line 38: maybe name the test "docker-composeinitial setup should have a rate of X". Just so we are clear this would a failure in docker-compose if were to happen

src/http/vault.ts
gregdhill: This functionality has been removed in the vault client. If required, we could add the calls for `updatePublicKey` and `registerAddress`.

src/http/vault.ts
gregdhill: Agreed. If `encodeBtcAddress` is the concern, just return an empty promise.

test/integration/apis/btc-core.test.ts
```diff
@@ -5,6 +5,12 @@ import { createPolkadotAPI } from "../../../src/factory";
 import { defaultEndpoint } from "../../config";
 import { BitcoinCoreClient } from "../../utils/bitcoin-core-client";
 
+// Test files are run alphabetically
+// but we want the oracle test to be run first
+// because it checks for the exchange rate, which
+// other tests later change
+require("./oracle.test.ts");
```
gregdhill, line 12: This makes testing in isolation tricky, I would probably prefer that tests don't require an explicit ordering to pass.

src/http/vault.ts
savudani8: I'm returning the JsonRpcResponse just in case we might need it in the future: https://github.com/interlay/polkabtc-js/pull/219/files#diff-923999cc29979ac83f22e93188908d26599e6cdaf298a00258ca5d2dffece334R81

src/http/vault.ts
savudani8: @nud3l Should I add `updatePublicKey` and `registerAddress` now or in a future PR?

src/http/vault.ts
gregdhill: Makes sense!

src/http/vault.ts
gregdhill: We would also need to add these calls to the `vault` client as well...

nud3l: Hm, easier of course if we can just remove this. How is this working on replace now @gregdhill? Do we derive an address for the vault as well or would the vault want to update their btc address inc ase they want to be replaced?

gregdhill: If the vault accepts a replace request it generates a new address via Bitcoin Core and submits that via the `accept_replace` call.

nud3l: Ok, then this is good to be removed since we don't need to update the address. @savudani8 can you please create a ticket on the UI board just to make sure that we remove that from the UI as well?

nud3l (approved):

PR #222, opened 2021-01-22
feat: Add faucet to docker-compose, enable integration tests

savudani8:


test/config.ts
```diff
@@ -1 +1,2 @@
-export const defaultEndpoint = "ws://0.0.0.0:9944";
+export const defaultParachainEndpoint = "ws://0.0.0.0:9944";
+export const defaultFaucetEndpoint = "http://0.0.0.0:3034";
```
gregdhill, line 2: Do we also want to add the default endpoints for the vault and staked-relayer?

savudani8: These are the only two endpoints that are being used for integration testing at the moment. I can add the others now, or when I actually write the tests for the other clients

gregdhill (approved):

PR #223, opened 2021-01-22
feat: Refund API

savudani8:


src/apis/refund.ts
```diff
@@ -0,0 +1,79 @@
...
+        return encodeRefundRequest(await this.api.query.refund.refundRequests(refundId), this.btcNetwork);
+    }
+
+    /**
+     * @param issueId The ID of the refund to fetch
+     * @returns A refund object
+     */
+    async getRequestByIssueId(issueId: string): Promise<RefundRequestExt> {
+        const customAPIRPC = this.api.rpc as any;
+        const keyValuePair = await customAPIRPC.refund.getRefundRequestsByIssueId(issueId);
```
nud3l, line 75: Since this call can return an error, this should be wrapped in a try catch block and the promise rejected in case there is none found or duplicates.

savudani8: Duplicates don't cause errors to be returned, because the parachain implementation simply returns the first Refund that matches the issue id

nud3l (approved):

PR #227, opened 2021-01-27
feat(transfer): Add transfer call on treasury

filip88:


src/apis/treasury.ts
```diff
@@ -24,4 +37,37 @@ export class DefaultTreasuryAPI implements TreasuryAPI {
...
+        }
+
+        return false;
+    }
+
+    /**
+     * @param destination The address of a user
+     * @param amount The Amount for a transfer
+     */
+    async transferPolkaBTC(destination: string, amount: string): Promise<void> {
```
nud3l, line 61: can you rename to "transfer" please?

src/apis/treasury.ts
```diff
@@ -24,4 +37,37 @@ export class DefaultTreasuryAPI implements TreasuryAPI {
...
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @param destination The address of a user
+     * @param amount The Amount for a transfer
```
nud3l, line 59: please rephrase: `@param amount - the amount in satoshi to transfer`

savudani8: Also renaming the parameter to `amountSatoshi` would be nice

src/apis/treasury.ts
```diff
@@ -1,13 +1,26 @@
 import { AccountId, Balance } from "@polkadot/types/interfaces/runtime";
 import { ApiPromise } from "@polkadot/api";
+import { sendLoggedTx } from "../utils";
+import { AddressOrPair } from "@polkadot/api/submittable/types";
+import { EventRecord } from "@polkadot/types/interfaces";
 
 export interface TreasuryAPI {
     totalPolkaBTC(): Promise<Balance>;
     balancePolkaBTC(id: AccountId): Promise<Balance>;
+    transferPolkaBTC(destination: string, amount: String): Promise<void>;
```
savudani8: I re-checked the type of the `api.tx.polkaBtc.transfer` and the second parameter (the `amount`) is of type `Compact<Balance> | AnyNumber | Uint8Array`, where `AnyNumber = BN | BigInt | Uint8Array | number | string;`. So the call should work with the lowercase `string` type

src/apis/treasury.ts
```diff
@@ -24,4 +37,37 @@ export class DefaultTreasuryAPI implements TreasuryAPI {
...
+        }
+
+        return false;
+    }
+
+    /**
+     * @param destination The address of a user
+     * @param amount The amount in satoshi to transfer
+     */
+    async transfer(destination: string, amount: string): Promise<void> {
```
nud3l, line 61: as Daniel mentioned, would be good to rename this to `amountSatoshi`

nud3l (approved):

savudani8:
The build is also failing because the mocked functions are empty. You can just use a `return;` statement to shut up the linter

nud3l (approved):

PR #233, opened 2021-02-02
feat(oracle-txfees): added API to set BTC tx fees

theodorebugnet:
Added get and set calls for `satoshi_per_bytes` and `set_btc_tx_fees_per_byte`.


src/apis/oracle.ts
```diff
@@ -18,12 +18,14 @@ export type OracleInfo = {
 
 export interface OracleAPI {
     getExchangeRate(): Promise<Big>;
+    getBtcTxFeesPerByte(): Promise<[number, number, number]>;
```
savudani8, line 21: You could return an object of type `BtcTxFeesPerByte` here

src/apis/oracle.ts
```diff
@@ -70,7 +72,42 @@ export class DefaultOracleAPI implements OracleAPI {
         }
         const encodedExchangeRate = encodeUnsignedFixedPoint(this.api, exchangeRate);
         const tx = this.api.tx.exchangeRateOracle.setExchangeRate(encodedExchangeRate);
-        const result = await sendLoggedTx(tx, this.account, this.api);
```
savudani8, line 73: This should not be removed. You can remove the left hand side of the assignment since it's not used, but we need the `await sendLoggedTx(tx, this.account, this.api);`

src/apis/oracle.ts
theodorebugnet: `BtcTxFeesPerByte` inherits from `Struct`; is that the best way to handle it?
I just changed it to a quick custom interface exported from apis/oracle.ts, let me know what you think. Seems easier to me since it can be constructed on the fly, as it doesn't inherit all the Struct members. It does introduce yet another type definition, but if the return of the RPC changes the lib will need updating anyway, so I don't feel like that's a problem.

If you disagree I'm happy to change it, though!

savudani8:
Nice! If can you also please increase the patch version in package.json please?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/polkabtc",
-  "version": "0.8.0",
+  "version": "0.8.1",
```
savudani8, line 3: This should be 0.8.2, as 0.8.1 has already been merged to `dev` and will probably be released before this commit

src/apis/oracle.ts
savudani8: That's a much better approach indeed!

package.json
theodorebugnet: Good call, gonna keep it in mind to pull dev before bumping a version

savudani8 (approved):

PR #238, opened 2021-02-10
feat: add getter for premium vaults

nud3l:


src/apis/vaults.ts
```diff
@@ -367,6 +368,18 @@ export class DefaultVaultsAPI {
         }
     }
 
+    async getPremiumRedeemVaults(): Promise<Map<AccountId, PolkaBTC>> {
+        const customAPIRPC = this.api.rpc as any;
```
savudani8, line 372: We can use actual types now that we updated to the latest `polkadot{.js}`, so no `any`

savudani8 (approved):

PR #248, opened 2021-02-19
Docs

savudani8:


docs/README.md
```diff
@@ -0,0 +1,108 @@
+@interlay/polkabtc / [Exports](/modules.md)
+
+# PolkaBTC JS
+
+JavaScript library to interact with PolkaBTC
+
+## Getting started
+
+See [BTC Parachain](https://github.com/interlay/btc-parachain) to get a development node running.
```
nud3l, line 9: I think the first instruction here should be the docker-compose since that's easier than to setup all the components individually.

docs/modules.md
```diff
@@ -0,0 +1,586 @@
...
+- [DefaultPolkaBTCAPI](/classes/defaultpolkabtcapi.md)
+- [FaucetClient](/classes/faucetclient.md)
+- [StakedRelayerClient](/classes/stakedrelayerclient.md)
+- [VaultClient](/classes/vaultclient.md)
+
+### Bitcoin Core Interfaces
+
+- [BTCCoreAPI](/interfaces/btccoreapi.md)
+
+### Other Interfaces
```
nud3l, line 22: Is it possible to rename this to "PolkaBTC Bridge Interfaces"?

nud3l (changes requested):
I think overall documentation (pending the two comments below) look great. Thanks for adding this!

I would like to change the way that they are generated though:
1. Instead of manually executing the `generate_docs` script, we add a GitHub action for this to run the script and then serve the docsify site. This prevents us from forgetting to update the docs and ensure we are always displaying the latest documentation alongside the code.
2. With the change in 1., please move the two files in the doc_helpers folder directly into the docs folder so that we just have everything in there.

savudani8:
> I think overall documentation (pending the two comments below) look great. Thanks for adding this!
> 
> I would like to change the way that they are generated though:
> 
>     1. Instead of manually executing the `generate_docs` script, we add a GitHub action for this to run the script and then serve the docsify site. This prevents us from forgetting to update the docs and ensure we are always displaying the latest documentation alongside the code.
> 
>     2. With the change in 1., please move the two files in the doc_helpers folder directly into the docs folder so that we just have everything in there.

The GitHub action is doable, but we cannot move the docs_helpers inside the docs directory. `typedoc` checks that either the `docs` directory does not exist or that that directory is in the exact structure it expects (and adding the docs_helpers breaks the structure).

nud3l:
I did not mean to copy the folder in there but the two files since they are duplicates right now. Does that work?

savudani8:
> I did not mean to copy the folder in there but the two files since they are duplicates right now. Does that work?

That doesn't work either. They're duplicated when the docs folder is generated, but the generation script removes this directroy as a first step so `typedoc` doesn't fail

nud3l:
> > I did not mean to copy the folder in there but the two files since they are duplicates right now. Does that work?
> 
> That doesn't work either. They're duplicated when the docs folder is generated, but the generation script removes this directory as a first step so `typedoc` doesn't fail

generate_docs > /tmp and then copy the files into the docs folder? It's just cleaner to have the doc related files in the doc folder.

nud3l (approved):

PR #249, opened 2021-02-23
Refactoring

savudani8:


src/index.ts
```diff
@@ -1,5 +1,5 @@
 export * from "./factory";
 export * from "./polkabtc-api";
-export * from "./apis";
-export * from "./http";
+export * from "./parachain";
+export * from "./clients";
 export * from "./utils";
```
nud3l, line 5: need to add `external` here I guess?

src/parachain/issue.ts
```diff
@@ -176,24 +170,19 @@ export class DefaultIssueAPI implements IssueAPI {
...
         const griefingCollateralPlanck = await this.getGriefingCollateralInPlanck(amountSat.toString());
-        const requestIssueTx = this.api.tx.issue.requestIssue(amountSat, vault.id, griefingCollateralPlanck);
+        const requestIssueTx = this.api.tx.issue.requestIssue(amountSat, vaultId, griefingCollateralPlanck);
         const result = await sendLoggedTx(requestIssueTx, this.account, this.api);
         if (!this.isRequestSuccessful(result.events)) {
             Promise.reject("Request failed");
         }
 
         const id = this.getIssueIdFromEvents(result.events);
         const issueRequest = await this.getRequestById(id);
```
nud3l, line 184: I checked the UI again, can you return the entire issue request here please? Because we are getting the id and then we are fetching the request right again from the parachain. So we can safe one `await` call.

src/parachain/redeem.ts
```diff
@@ -199,7 +193,7 @@ export class DefaultRedeemAPI {
         if (!this.isRequestSuccessful(result.events)) {
             throw new Error("Request failed");
         }
-        const id = this.getRedeemIdFromEvents(result.events, "RequestRedeem");
+        const id = this.getRedeemIdFromEvents(result.events, this.api.events.redeem.RequestRedeem);
         return { id, vault };
```
nud3l, line 197: Here as well, please return the redeem request. From this, we have all the info we need in the UI. I think it's also the logical thing to return from the request call.

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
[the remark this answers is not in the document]
savudani8: We do, it stores the husky logic for applying pre-commit hooks

docker-compose.yml
```diff
@@ -1,11 +1,13 @@
 version: "3.8"
 services:
     polkabtc:
-        image: "registry.gitlab.com/interlay/btc-parachain:0-4-0"
+        image: "registry.gitlab.com/interlay/btc-parachain:0-5-0"
```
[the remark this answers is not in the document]
savudani8: Updated to parachain 0.5.3 and clients 0.5.6

src/parachain/redeem.ts
```diff
@@ -153,14 +153,11 @@ export class DefaultRedeemAPI {
      * @returns A boolean value
      */
     isRequestSuccessful(events: EventRecord[]): boolean {
```
nud3l, line 155: I think eventually, these functions can be all moved into one. Since now you have typed events, it can just take the event vector plus the event you are searching for as input. What do you think? Can be done in a follow-up PR though! This one is already very large.

nud3l (approved):

src/parachain/redeem.ts
savudani8: That's a great idea! Something similar is already being done here: https://github.com/interlay/polkabtc-js/blob/dan/refactoring/src/parachain/redeem.ts#L133

PR #250, opened 2021-02-26
feat: Get btc tx by id

savudani8:


src/apis/btc-core.ts
```diff
@@ -142,6 +165,27 @@ export class DefaultBTCCoreAPI implements BTCCoreAPI {
         return this.getData(this.txApi.getTxMerkleBlockProof(txid));
     }
 
+    getTx(txid: string): Promise<Transaction> {
+        return this.getData(this.txApi.getTx(txid));
+    }
+
+    async getUtxoAmount(txid: string, recipient: string): Promise<string> {
+        let amount = "0";
```
gregdhill, line 173: this should be a `number`

nud3l: is `vout.value` a `number`? Just checking since it might be a string due to the limited `number` size

savudani8: It is a number

nud3l (approved):

PR #257, opened 2021-03-10
Correct BTC confirmations count

theodorebugnet:
Changes the confirmations calculation for BTC transactions to match Bitcoin convention of calculating them as block depth.


src/external/btc-core.ts
```diff
@@ -284,8 +284,12 @@ export class DefaultBTCCoreAPI implements BTCCoreAPI {
         status.confirmed = txStatus.confirmed;
         // NOTE: the second part of the check is only to ensure that the -1 we sometimes see
         // in the UI is not caused by the lib
-        if (txStatus.block_height && latest_block_height - txStatus.block_height > 0) {
-            status.confirmations = latest_block_height - txStatus.block_height;
+        if (txStatus.block_height && latest_block_height - txStatus.block_height >= 0) {
```
savudani8, line 287: > note that block_height will only be set if confirmed == true
Would it be cleaner to replace `txStatus.block_height` in the `if` condition with `confirmed == true` then?

savudani8 (approved):

src/external/btc-core.ts
theodorebugnet: These should be equivalent (https://github.com/Blockstream/electrs/blob/new-index/src/util/transaction.rs#L28) but you're right from a readability point of view.

theodorebugnet: Never mind, that trips up the typescript type checker. I'll leave it as is.

PR #258, opened 2021-03-10
feat: Add core vault and relayer RPCs to API, refactor fees and trans…

savudani8:
…action correctness

Doesn't remove the `vault` and `relayer` from the `clients` directory yet, because only the functionality used in the UI has been transitioned to the API. The transition will be finished in a future PR


src/parachain/fee.ts
```diff
@@ -0,0 +1,74 @@
...
+     * @param feesDOT Planck value representing the DOT fees accrued
+     * @param lockedDOT Planck value representing the value locked to gain yield
+     * @param dotToBtcRate Conversion rate
+     * @returns The APY, given the parameters
+     */
+    calculateAPY(feesPolkaBTC: string, feesDOT: string, lockedDOT: string, dotToBtcRate: Big): string;
+    /**
+     * @returns The griefing collateral rate for issuing PolkaBTC
+     */
+    getIssueGriefingCollateralRate(): Promise<UnsignedFixedPoint>;
```
nud3l, line 33: I think this should return a string or a `Big`. I think the conversion should happen inside the lib.

src/parachain/fee.ts
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * @param amountSat Amount, in Satoshi, for which to compute the required
+     * griefing collateral, in Planck
+     * @param griefingCollateralRate
+     * @param oracleAPI
+     * @returns The griefing collateral, in Planck
+     */
+     getGriefingCollateralInPlanck(
+        amountSat: PolkaBTC,
+        griefingCollateralRate: UnsignedFixedPoint
```
nud3l, line 20: Should be input as a string or Big here

src/parachain/fee.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    constructor(private api: ApiPromise) {
+        this.oracleAPI = new DefaultOracleAPI(api);
+    }
+
+    async getGriefingCollateralInPlanck(
+        amountSat: PolkaBTC,
+        griefingCollateralRate: UnsignedFixedPoint
+    ): Promise<Big> {
+        const griefingCollateralRateBig = new Big(decodeFixedPointType(griefingCollateralRate));
```
nud3l, line 47: by having a `Big` for griefing collateral rate, we prevent from calling this on every update.

src/parachain/fee.ts
```diff
@@ -0,0 +1,74 @@
...
+        const planckPerSatoshi = await this.oracleAPI.getRawExchangeRate();
```
nud3l, line 48: Would be worth adding a conversion function to the oracle to get from Sat to Planck without re-implementing this conversion in other places.

src/parachain/fee.ts
```diff
@@ -0,0 +1,74 @@
...
+        const griefingCollateralPlanckRoundedUp = new Big(griefingCollateralPlanck).round(0, 3);
+        return griefingCollateralPlanckRoundedUp;
+    }
+
+    async getIssueGriefingCollateralRate(): Promise<UnsignedFixedPoint> {
+        return await this.api.query.fee.issueGriefingCollateral();
+    }
+
+    calculateAPY(feesPolkaBTC: string, feesDOT: string, lockedDOT: string, dotToBtcRate: Big): string {
+        const feesPolkaBTCBig = new Big(feesPolkaBTC.toString());
```
nud3l, line 65: Already a string, why the second conversion?

src/parachain/issue.ts
```diff
@@ -175,25 +136,20 @@ export class DefaultIssueAPI implements IssueAPI {
...
 
-    async execute(issueId: H256, txId: H256Le, merkleProof: Bytes, rawTx: Bytes): Promise<boolean> {
+    async execute(issueId: H256, txId: H256Le, merkleProof: Bytes, rawTx: Bytes): Promise<void> {
         if (!this.account) {
             throw new Error("cannot request without setting account");
         }
         const executeIssueTx = this.api.tx.issue.executeIssue(issueId, txId, merkleProof, rawTx);
-        const result = await this.transaction.sendLogged(executeIssueTx, this.account);
-        return this.isExecutionSuccessful(result.events);
+        await this.transaction.sendLogged(executeIssueTx, this.account, this.api.events.issue.ExecuteIssue);
```
nud3l, line 152: Is the Promise rejected if unsuccessful?

src/parachain/redeem.ts
```diff
@@ -200,29 +163,21 @@ export class DefaultRedeemAPI {
             throw new Error("cannot execute without setting account");
         }
         const executeRedeemTx = this.api.tx.redeem.executeRedeem(redeemId, txId, merkleProof, rawTx);
-        const result = await this.transaction.sendLogged(executeRedeemTx, this.account);
-        if (!this.isExecutionSuccessful(result.events)) {
-            throw new Error("Execution failed");
-        }
+        const result = await this.transaction.sendLogged(executeRedeemTx, this.account, this.api.events.redeem.ExecuteRedeem);
```
nud3l, line 166: I like this new checking, much more elegant than the way we had it before!

src/utils/transaction.ts
```diff
@@ -22,9 +24,10 @@ export interface TransactionAPI {
 export class Transaction implements TransactionAPI {
     constructor(private api: ApiPromise) {}
 
-    async sendLogged(
+    async sendLogged<T extends AnyTuple>(
         transaction: SubmittableExtrinsic<"promise">,
-        signer: AddressOrPair
+        signer: AddressOrPair,
+        successEventType?: AugmentedEvent<ApiTypes, T>
```
nud3l, line 30: Can you please add the parameters in the interface doc comment?

src/parachain/issue.ts
savudani8: It is rejected here to avoid duplication: https://github.com/interlay/polkabtc-js/blob/dan/client-transition-to-api/src/utils/transaction.ts#L63

nud3l (approved):

nud3l:
LGTM!

PR #262, opened 2021-03-16
feat: add fee rate to issue API

nud3l:


src/parachain/redeem.ts
```diff
@@ -227,16 +227,16 @@ export class DefaultRedeemAPI {
     }
 
     async getFeesToPay(amount: string): Promise<string> {
-        const feePercentage = await this.getFeePercentage();
+        const feePercentage = await this.getFeeRate();
         const feePercentageBN = new Big(feePercentage);
```
savudani8, line 231: We can remove this line, since the `getRate()` function returns Big already (the same way you did for issue)

nud3l: Done!

savudani8 (approved):

PR #272, opened 2021-03-24
[DRAFT] Request aggregator

theodorebugnet:
Generally complete, but waiting until the lib becomes compatible with the latest parachain changes (will need updating `polkabtc-types` dependency and regenerating parachain.json).

Old notes/discussion:
 - **Atomicity**: a parameter has been added to configure whether a batch is atomic or not (true by default).
 - **Vault selection algorithm**: I implemented a semi-naive method which greedily selects top vaults if no single vault can fulfil a request, but selects at random if more than one vault can fulfil a request (or the remainder of a partially-allocated aggregated request). This is less evenly distributed than it could be (e.g. there may be multiple allocations which use the same number of vaults, but the greedy approach will always find the same one; or maybe a 50%-50% distribution between two vaults leaving capacity in both could be better than filling the first one and allocating only a small remainder to a second one) but I think it's good enough for now and relatively simple.
Additionally a `retries` parameter has been added, to re-run the selection up to n times in case the entire request fails to be filled on the first try.


savudani8:
> Atomicity

We could have a boolean parameter that is false by default and denotes the atomicity. Depending on the number of vaults involved, it seems likely that a vault will have less issuable tokens between the `getRequestLimits` and the `request` calls. This could be represented as a checkmark in the UI.

> UI Flow / Caching

That would be a bit of an unusual pattern indeed. How about having `vaults` as an optional parameter in both `getRequestLimits` and `request`? The UI would fetch the vaults once and pass the same input to the functions.

> Vault selection algorithm

It seems excellent for now. Further on maybe we'd prioritise vaults based on SLA. The fact that the algorithm tends to use the same vaults can be tied to the atomicity issue above

src/parachain/issue.ts
```diff
@@ -113,40 +140,100 @@ export class DefaultIssueAPI implements IssueAPI {
         this.transaction = new Transaction(api);
     }
 
+    async getRequestLimits(): Promise<IssueLimits> {
+        const vaults = await this.vaultsAPI.getVaultsWithIssuableTokens();
+        const [singleVault, maxTotal] = [...vaults.entries()].reduce(([singleVault, maxTotal], [_, vaultAvailable]) => {
```
savudani8, line 145: If I understand this reducer correctly, more suggestive would be `[singleVaultMaxIssuable, totalIssuable]` instead of `[singleVault, maxTotal]`

src/parachain/issue.ts
```diff
@@ -113,40 +140,100 @@ export class DefaultIssueAPI implements IssueAPI {
...
+        amountsPerVault: Map<AccountId, PolkaBTC>,
+        griefingCollateralRate: Big
+    ): Promise<IssueRequestResult[]> {
+        if (!this.account) {
+            return Promise.reject(ACCOUNT_NOT_SET_ERROR_MESSAGE);
+        }
 
+        const txes = new Array<SubmittableExtrinsic<"promise">>();
+        for (const [vault, amount] of amountsPerVault) {
+            const griefingCollateral = await this.feeAPI.getGriefingCollateralInPlanck(amount, griefingCollateralRate);
```
savudani8, line 225: This makes me wish we had implemented some batching in the `Transaction` API

nud3l:
> WIP on the issue side for now - opening for review before proceeding to change tests and implement redeem.
> 
> Some notes:
> 
>     * **Atomicity**: a batch of issue transactions is currently atomic; if you make an issue request that would require e.g. 10 vaults, and a single issue request fails (e.g. someone else makes an issue request with one of those vaults, reserves their collateral, and they can no longer fulfil yours) the entire batch fails. This can be trivially changed to be non-atomic (`batchAll` vs `batch`); I'm not sure which way is better.
> 
>     * **Flow with the UI.** To calculate the thresholds for issue across one or all vaults, the list of vaults must be fetched. To then perform the issue requests, the list of vaults must be fetched. As the lib is stateless, I added the option for the UI to cache the list of vaults and pass it back, which slightly reduces the vision of an elegant single-parameter `request(amount)`, but saves an extra call to the parachain.
> 
>     * **Vault selection algorithm**: I implemented a semi-naive method which greedily selects top vaults if no single vault can fulfil a request, but selects at random if more than one vault can fulfil a request (or the remainder of an aggregated request). This is less evenly distributed than it could be (e.g. there may be multiple allocations which use the same number of vaults, but the greedy approach will always find the same one; or maybe a 50%-50% distribution between two vaults leaving capacity in both could be better than filling the first one and allocating only a small remainder to a second one) but I think it's good enough for now and relatively simple.

Sorry for taking so long to answer!

It's overall probably better to do atomic issuing, but I like Dan's idea of having a boolean. Further to that, you could also add a "retry" parameter. Thinking about the list of vaults that you are passing and the selection algorithm: if there are enough vaults available, the user could pass that retry parameter to e.g. try at least three more vaults if the first set we thought was viable does not work. In that case, if e.g. just a fraction of the PolkaBTC is missing, the lib could just try filling this with other vaults. Since you are using the greedy algorithm (which I think is great for now), users that have the same list will likely select the same vaults so there is a decent probability that collisions will happen. What do you think?

src/parachain/issue.ts
```diff
@@ -34,13 +36,33 @@ export function encodeIssueRequest(req: IssueRequest, network: Network): IssueRe
...
+    request(amountSat: PolkaBTC, availableVaults?: Map<AccountId, PolkaBTC>): Promise<IssueRequestResult[]>;
+
+    /**
+     * Send a batch of aggregated issue transactions (to one or more vaults)
+     * @param amountSat PolkaBTC amounts (denoted in Satoshi) to issue using each vault
+     * @param vaultId The vaults from which to request issue
+     * @returns An array of type {issueId, vault, error} if the requests succeeded. error will be null for successful
+     * requests; issueId and vault will be null for failed ones.
+     */
+    requestAdvanced(
```
nud3l, line 61: maybe a better name would be `batchRequest`?

theodorebugnet:
> It's overall probably better to do atomic issuing, but I like Dan's idea of having a boolean. Further to that, you could also add a "retry" parameter.
> Since you are using the greedy algorithm (which I think is great for now), users that have the same list will likely select the same vaults so there is a decent probability that collisions will happen. What do you think?
I think that with 6 second block times (in theory), collisions between multiple users trying to batch-issue all at once don't seem very likely. If two users do collide, the next one can just promptly re-request within a few seconds.
For small requests (which are likely going to be much more frequent than huge ones, I'm guessing), I'm thinking leaving a portion of vaults free would help. E.g. if the greedy algorithm reserves 80% of each vault, then people can still request up to 20% in small requests without blocking the large batch. Downside being that this might end up using more vaults than 100% utilisation - marginally increasing bitcoin fees (though I'm not sure how much a couple thousand sats matters if you're already making a multi-btc request).

Those are my general thoughts on blocking. Retry is definitely a possible idea as well, though if the transaction is retried many times it might significantly delay the UI. I also think that if the algorithm still reserves 100% of each vault every time it retries, in a very busy system (where the volume of issue requests per block is proportional to the number of operating vaults) there's the potential to get nowhere even with multiple retries.

src/parachain/issue.ts
theodorebugnet: My reasoning behind this is that this is basically what would be called if someone wants to make sure to manually pass every parameter. It doesn't have to be used for batching, either - if you want to request with just a single vault, but you want to specify exactly which vault to use, you'd call `requestAdvanced` rather than `request`. E.g. for the power-user mode you mentioned.

Conversely, you can also batch request by calling `request` (it just auto-selects the vaults for you, if necessary).

src/parachain/issue.ts
```diff
@@ -34,13 +36,47 @@ export function encodeIssueRequest(req: IssueRequest, network: Network): IssueRe
...
+     * @param amountSat PolkaBTC amount (denoted in Satoshi) to issue.
+     * @param availableVaults (optional) A list of all vaults usable for issue. If not provided, will fetch from the
+     * parachain (incurring an extra request).
+     * @param atomic (optional) Whether the issue request should be handled atomically or not. Only makes a difference
+     * if more than one vault is needed to fulfil it. Defaults to false.
+     * @returns An array of type {issueId, vault, error} if the requests succeeded. error will be null for successful
+     * requests; issueId and vault will be null for failed ones.
+     */
+    request(
+        amountSat: PolkaBTC,
```
nud3l, line 58: btw, while you are here: we should refactor this to use `BN`

src/parachain/issue.ts
```diff
@@ -213,8 +342,16 @@ export class DefaultIssueAPI implements IssueAPI {
     }
 
     async getRequestById(issueId: H256): Promise<IssueRequestExt> {
+        return (await this.getRequestsByIds([issueId]))[0];
```
nud3l, line 345: can you add in the head part to only take this from the finalized one?

src/parachain/issue.ts
```diff
@@ -113,40 +154,128 @@ export class DefaultIssueAPI implements IssueAPI {
...
+                // select random vault able to fulfil request
+                const range = vaultsArray.length - firstSuitable;
+                const idx = Math.floor(Math.random() * range) + firstSuitable;
+                console.log(`Selecting vault at index ${idx}`);
+                vault = vaultsArray[idx][0];
+                amount = amountToAllocate;
+                console.log(`Allocating ${amount} to ${vault} all at once`);
+            } else {
+                // else allocate greedily
+                if (vaultsArray.length === 0) throw new Error("Insufficient capacity to fulfil request");
```
nud3l, line 213: that should likely be thrown earlier or the check here should be if length < 2 or so

theodorebugnet: Good idea, I'll add another check at the start.
I'll still keep this one in to make the type safety obvious.

src/parachain/issue.ts
theodorebugnet: Not sure what you mean - do you mean in the doc comment? What do you mean by "only take this from the finalized one"?
(In case it's not clear, it's the same functionality and logic as it always has been, it just now works by wrapping an array-capable version (to avoid duplicating the same logic in both).)

nud3l: ah, nevermind. I missed that it calls the function below.

nud3l:
> I think that with 6 second block times (in theory), collisions between multiple users trying to batch-issue all at once don't seem very likely. If two users do collide, the next one can just promptly re-request within a few second

Yes and no. The thing is that the list of vaults is loaded once the app is loading and not updated every 6 seconds. So if you have multiple users connecting and no users have made an issue request for some time, they will see the same state. It would only be updated if they refresh the page.

> Those are my general thoughts on blocking. Retry is definitely a possible idea as well, though if the transaction is retried many times it might significantly delay the UI. I also think that if the algorithm still reserves 100% of each vault every time it retries, in a very busy system (where the volume of issue requests per block is proportional to the number of operating vaults) there's the potential to get nowhere even with multiple retries.

Agree, the default retry strategy should be 0. Can even have the parameter like `retry = 0` so that you don't have to set it in the UI.

src/parachain/issue.ts
gregdhill: Hmm, I'm not thrilled with this naming since it doesn't better describe the difference between this and the other `request` method.

src/parachain/redeem.ts
```diff
@@ -49,10 +48,32 @@ export interface RedeemAPI extends TransactionAPI {
     /**
      * Send a redeem request transaction
      * @param amount PolkaBTC amount (denoted in Bitcoin) to redeem
+>>>>>>> 54d4ba92bf8213f6bda540fd4851b5ea46e6fd84
```
gregdhill, line 51: remove

src/parachain/issue.ts
```diff
@@ -121,33 +159,87 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
...
+        }
+    ): Promise<IssueRequestResult[]> {
+        try {
+            const availableVaults = options?.availableVaults || await this.vaultsAPI.getVaultsWithIssuableTokens();
+            const atomic = !!options?.atomic;
+            const retries = options?.retries || 0;
+            console.log("Available vaults on ISSUE:");
+            [...availableVaults.entries()].map(([vaultId, amount]) => {
+                console.log(`${amount.toString()} available with ${vaultId.toString()}`);
+            });
```
gregdhill, line 199: remove debug logging

src/parachain/issue.ts
```diff
@@ -121,33 +159,87 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
...
+                amountSat: ${amountWrapped.toString()}
+                , to vaultId: ${vault.toString()}
+                , with griefing collateral: ${griefingCollateral.toString()}`);
+            txes.push(this.api.tx.issue.requestIssue(amountWrapped, vault, dotToPlanck(griefingCollateral.toString()) as string));
         }
-        const amountSat = this.api.createType("Wrapped", btcToSat(amount.toString()));
-        const griefingCollateral = await this.getGriefingCollateral(amount);
-        const requestIssueTx = this.api.tx.issue.requestIssue(amountSat, vaultId, dotToPlanck(griefingCollateral.toString()) as string);
-        const result = await this.sendLogged(requestIssueTx, this.api.events.issue.RequestIssue);
+        const batch = (atomic ? this.api.tx.utility.batchAll : this.api.tx.utility.batch)(txes);
```
gregdhill, line 235: What's the difference between `batchAll` and `batch`?

gregdhill: Maybe add a comment here

src/parachain/issue.ts
theodorebugnet: Definitely, this is just WIP while I'm debugging tests

theodorebugnet: Definitely, this is just WIP while I'm debugging tests

theodorebugnet: Definitely, this is just WIP while I'm debugging tests

src/parachain/issue.ts
```diff
@@ -218,24 +218,15 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
...
+        const txs = new Array<SubmittableExtrinsic<"promise">>();
         for (const [vault, amount] of amountsPerVault) {
-            const griefingCollateral = await this.getGriefingCollateral(amount);
-            const amountWrapped = this.api.createType("Wrapped", btcToSat(amount.toString()));
-            console.log(`Issuing:
-                - amountSat: ${amountWrapped.toString()}
-                - to vaultId: ${vault.toString()}
-                - with griefing collateral: ${griefingCollateral.toString()}`);
-            txes.push(this.api.tx.issue.requestIssue(amountWrapped, vault, dotToPlanck(griefingCollateral.toString()) as string));
+            const griefingCollateral = (await this.getGriefingCollateral(amount)).add(1);
```
theodorebugnet, line 223: Quick question, why `add(1)`? Especially since this is still the DOT value, this looks like it will inflate all griefing collaterals by 1 DOT.

src/parachain/issue.ts
theodorebugnet: Would you have a better suggestion? In my mind the main difference is that `request` auto-selects everything with just an amount, while `requestAdvanced` requires specifying what you want to request and where and doesn't set any parameters itself.

In the past there was also griefing rate, now that it's removed maybe `requestWithVaultList` is a bit more accurate, but I don't like that name. Perhaps something like `requestManually`.

src/parachain/issue.ts
gregdhill: Yeah this was a quick hack, there is some discrepancy between this calculation of the griefing collateral and that on-chain. Likely related to the lose of precision going to and from Planck - DOT.

theodorebugnet: Have you tested if your rounding up changes fix that - or should we add a few Planck instead?

src/parachain/issue.ts
```diff
@@ -121,33 +159,73 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
         this.feeAPI = new DefaultFeeAPI(api);
     }
 
+    async getRequestLimits(vaults?: Map<AccountId, Big>): Promise<IssueLimits> {
+        if (!vaults) vaults = await this.vaultsAPI.getVaultsWithIssuableTokens();
+        const [singleVaultMaxIssuable, totalMaxIssuable] = [...vaults.entries()].reduce(
+            ([singleVault, maxTotal], [_, vaultAvailable]) => {
+                maxTotal = maxTotal.plus(vaultAvailable);
+                singleVault = singleVault.gt(vaultAvailable) ? singleVault : vaultAvailable;
```
nud3l, line 167: The `getVaultsWithIssuableTokens()` is sorted so to get the maximum for a single vault, you can just take the top element.

src/parachain/issue.ts
```diff
@@ -121,33 +159,73 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
...
-            if (this.api.events.issue.RequestIssue.is(event)) {
-                const hash = this.api.createType("Hash", event.data[0]);
-                return hash;
+    private getIssueIdsFromEvents(events: EventRecord[]): Hash[] {
+        return getRequestIdsFromEvents(events, this.api.events.issue.RequestIssue, this.api);
+    }
+
+    async request(
+        amount: Big,
+        options?: {
```
nud3l, line 186: I think in most places we flattened the options. Any particular reason you want to submit an options object here?

src/parachain/issue.ts
```diff
@@ -121,33 +159,73 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
...
-            vaultId = await this.vaultsAPI.selectRandomVaultIssue(amount);
+    async requestAdvanced(
+        amountsPerVault: Map<AccountId, Big>,
+        atomic: boolean
+    ): Promise<IssueRequestResult[]> {
+        const txs = new Array<SubmittableExtrinsic<"promise">>();
+        for (const [vault, amount] of amountsPerVault) {
+            const griefingCollateral = await this.getGriefingCollateral(amount);
+            // mul() here is a hacky workaround for rounding errors
+            const griefingCollateralPlanck = new Big(dotToPlanck(griefingCollateral.toString()) || "0").mul(1.01);
```
nud3l, line 217: What is happening here? It's basically off by a couple of planck or why increase this by 1%?

src/parachain/issue.ts
theodorebugnet: Mainly because there's no definite order for the optional arguments. But I think they can actually be flattened by putting the `availableVaults` argument last, that should work for most cases.

src/parachain/issue.ts
theodorebugnet: Yep, you can check the discussion in discord - the request fails with `InsufficientCollateral` without this, despite best attempts to round the plank amount upwards.

theodorebugnet: To add, I'm testing whether this can be lowered. If it works when I add just 1-2 Planck rather than a percentage, I'll do that instead.

src/parachain/issue.ts
```diff
@@ -121,33 +160,69 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
         this.feeAPI = new DefaultFeeAPI(api);
     }
 
+    async getRequestLimits(vaults?: Map<AccountId, Big>): Promise<IssueLimits> {
+        if (!vaults) vaults = await this.vaultsAPI.getVaultsWithIssuableTokens();
+        const vaultsArr = [...vaults.entries()];
+        const singleVaultMaxIssuable = vaultsArr[0][1];
```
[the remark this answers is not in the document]
theodorebugnet: > Why
The PRC call returns a sorted list, as pointed out by Dom.
> What happens when the array is empty
Good point. Result should be 0 but this will throw. Will fix.

src/parachain/issue.ts
```diff
@@ -121,33 +160,69 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
...
-            vaultId = await this.vaultsAPI.selectRandomVaultIssue(amount);
+    async requestAdvanced(
+        amountsPerVault: Map<AccountId, Big>,
+        atomic: boolean
+    ): Promise<IssueRequestResult[]> {
+        const txs = new Array<SubmittableExtrinsic<"promise">>();
+        for (const [vault, amount] of amountsPerVault) {
+            const griefingCollateral = await this.getGriefingCollateral(amount);
+            // mul() here is a hacky workaround for rounding errors
+            const griefingCollateralPlanck = new Big(dotToPlanck(griefingCollateral.toString()) || "0").add(100);
```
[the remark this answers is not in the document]
theodorebugnet: I tried testing, 10 Planck didn't work, 100 did. I could try bisecting further.

src/parachain/redeem.ts
```diff
@@ -207,8 +248,11 @@ export class DefaultRedeemAPI extends DefaultTransactionAPI implements RedeemAPI
         [merkleProof, rawTx] = await getTxProof(this.electrsAPI, btcTxId, merkleProof, rawTx);
         const executeRedeemTx = this.api.tx.redeem.executeRedeem(parsedRequestId, merkleProof, rawTx);
         const result = await this.sendLogged(executeRedeemTx, this.api.events.redeem.ExecuteRedeem);
-        const id = this.getRedeemIdFromEvents(result.events, this.api.events.redeem.ExecuteRedeem);
-        if (id) {
+        const ids = this.getRedeemIdsFromEvents(result.events);
+        if (ids.length > 1) {
+            throw new Error("Unexpected multiple redeem events from single execute transaction!");
```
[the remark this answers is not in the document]
theodorebugnet: Yep, these are events returned by the specific extrinsic dispatched as far as I understand it.

src/utils/issueRedeem.ts
```diff
@@ -0,0 +1,74 @@
...
+import type { AnyTuple } from "@polkadot/types/types";
+import { ApiPromise } from "@polkadot/api";
+
+/**
+ * @param events The EventRecord array returned after sending a transaction
+ * @param methodToCheck The name of the event method whose existence to check
+ * @returns The id associated with the transaction. If the EventRecord array does not
+ * contain required events, the function throws an error.
+ */
+export function getRequestIdsFromEvents(
```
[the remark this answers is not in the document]
theodorebugnet: The events are filtered by event type, so e.g. `RequestIssue` events only.

PR #276, opened 2021-03-26
feat(oracle): Return names mapped by id

savudani8:


src/parachain/oracle.ts
```diff
@@ -53,17 +46,13 @@ export interface OracleAPI {
      */
     getLastExchangeRateTime(): Promise<Date>;
     /**
-     * @returns An array with the oracle names
+     * @returns A map from the oracle's account id to its
```
theodorebugnet, line 49: ...`name`

src/parachain/oracle.ts
```diff
@@ -89,6 +78,11 @@ export interface OracleAPI {
      * @returns Convert a Satoshi amount to Planck
      */
     convertSatoshiToPlanck(satoshi: PolkaBTC): Promise<Big>;
+    /**
+     * @returns The period of time (in milliseconds) after an oracle's last submission
+     * during which it is considered online
+     */
+    getMaxDelay(): Promise<number>;
```
theodorebugnet, line 85: I would name this something like `getOnlineTimeout`, what do you think?

theodorebugnet (approved):

src/parachain/oracle.ts
nud3l: it's in line with the name on the chain, so I think fine to keep

savudani8: I agree with @theodorebugnet that it's more suggestive. We could change the naming in the parachain

nud3l: ok, agree!

PR #277, opened 2021-03-31
feat(redeem): Liquidation (burn) redeem

savudani8:
It takes about 15 mins for the staked relayer to report the theft, which makes the integration tests take a while


nud3l:
> It takes about 15 mins for the staked relayer to report the theft, which makes the integration tests take a while

Do you know if there is a way to "classify" tests easily? I'm thinking about having certain tests tagged as release tests that we only ever want to execute when we tag a new release. Because waiting 15 min on every pipeline is not really something any of us would want.

The straightforward way would of course be to have the tests in a separate folder like the unit and integration tests. But technically this is also an integration test so logically it Fotos be in the same folder.

src/parachain/collateral.ts
```diff
@@ -15,17 +15,17 @@ export interface CollateralAPI {
     /**
      * @returns Total locked DOT collateral
      */
-    totalLockedDOT(): Promise<Balance>;
+    totalLockedDOT(): Promise<BN>;
```
nud3l, line 18: I'd suggest we remove the DOT part in the name since this will change on Kusama and this is anyway a breaking change for the UI.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/polkabtc",
-  "version": "0.11.2",
+  "version": "0.11.3",
```
nud3l, line 3: Should be minor release due to breaking changes

src/parachain/redeem.ts
```diff
@@ -118,16 +123,33 @@ export interface RedeemAPI {
...
+     * @param amount The amount of PolkaBTC to burn, denominated as PolkaBTC
+     */
+    burn(amount: Big): Promise<void>;
+    /**
+     * @returns The maximum amount of tokens that can be burned through a liquidation redeem
+     */
+    getMaxBurnableTokens(): Promise<Big>;
+    /**
+     * @returns The exchange rate (collateral currency to wrapped token currency)
+     * used when burning tokens. This exchange rate is at least as favourable to the
```
nud3l, line 137: Actually that is not always the case. In massive dot exchange rate crashes we might end up with lower dot amounts than you get for burning polkabtc

src/parachain/redeem.ts
```diff
@@ -118,16 +123,33 @@ export interface RedeemAPI {
      * and required completion time by a user.
      */
     getRedeemPeriod(): Promise<BlockNumber>;
+    /**
+     * Send a transaction to set the DOT/BTC exchange rate
```
nud3l, line 127: Seems like a typo? This didn't change the rate but burns polkabtc for dot

src/parachain/redeem.ts
```diff
@@ -187,6 +209,32 @@ export class DefaultRedeemAPI {
         await this.transaction.sendLogged(cancelRedeemTx, this.account, this.api.events.redeem.CancelRedeem);
     }
 
+    async burn(amount: Big): Promise<void> {
+        if (!this.account) {
+            return Promise.reject(ACCOUNT_NOT_SET_ERROR_MESSAGE);
+        }
+        const amountSat = this.api.createType("Balance", btcToSat(amount.toString()));
+        const cancelRedeemTx = this.api.tx.redeem.liquidationRedeem(amountSat);
+        await this.transaction.sendLogged(cancelRedeemTx, this.account, this.api.events.redeem.LiquidationRedeem);
```
nud3l, line 218: This API was a really nice addition!

src/parachain/vaults.ts
```diff
@@ -362,6 +370,18 @@ export class DefaultVaultsAPI {
         return encodeVault(vault, this.btcNetwork);
     }
 
+    async getLiquidationVaultId(): Promise<string> {
```
nud3l, line 373: Do you need that API since it does the same as the call below?

savudani8: It's more lightweight than the call below, so I thought it might be useful

savudani8:
> > It takes about 15 mins for the staked relayer to report the theft, which makes the integration tests take a while
> 
> Do you know if there is a way to "classify" tests easily? I'm thinking about having certain tests tagged as release tests that we only ever want to execute when we tag a new release. Because waiting 15 min on every pipeline is not really something any of us would want.
> 
> The straightforward way would of course be to have the tests in a separate folder like the unit and integration tests. But technically this is also an integration test so logically it Fotos be in the same folder.

We could do this via the workflow actions. First, we need to suffix test names as a tag, to achieve this separation. In `package.json`:
- "test:integration:staging": "mocha test/integration/**/*.staging.test.ts
- "test:integration:release": "mocha test/integration/**/*.release.test.ts
This means we'll have two versions of [test.yml](https://github.com/interlay/polkabtc-js/blob/master/.github/workflows/test.yml), one triggered by a push to the master branch and one triggered by PRs.
Then, [npm.yml](https://github.com/interlay/polkabtc-js/blob/master/.github/workflows/npm.yml) will depend on `test-master.yml` to succeed. A problem here is that if a tag is pushed without `test-master.yml` being successful, the `npm.yml`  action will have to be restarted manually. This is to avoid needlessly pushing tags to trigger the npm release script.

@theodorebugnet What do you think?

nud3l:
I think one way around that issue is that on successful completion of the release pipeline, the pipeline can just tag the version and thereby trigger the npm pipeline. That would also resolve the manual step of us tagging. I also don't really see a use case wehre we want to update master without a new release.

@ns212 I think that should be possible with jenkins?

We can do the same for stats btw. What do you guys think?

nud3l:
Pending the testing/pipeline change this LGTM.

savudani8:
@nud3l How would the auto-tagging work though? You mentioned a tool a while ago that parses commit messages to do semantic versioning. Perhaps this one: https://github.com/semantic-release/semantic-release

savudani8:
We cannot enforce `release` integration tests as a dependency of the `npm` action, because they are part of different workflow files. So for now, tagging must happen after manually ensuring that the `release` pipeline succeeded. 

I removed the integration test that checks for successful provider disconnection, because the parachain doesn't process it quickly enough, causing it to fail (here is the provider logic https://github.com/polkadot-js/api/blob/master/packages/rpc-provider/src/ws/index.ts#L355). Until now, that test was left out of the `integration` test directory and was never run.

nud3l:
> We cannot enforce `release` integration tests as a dependency of the `npm` action, because they are part of different workflow files. So for now, tagging must happen after manually ensuring that the `release` pipeline succeeded. 
> 
> 
> 
> I removed the integration test that checks for successful provider disconnection, because the parachain doesn't process it quickly enough, causing it to fail (here is the provider logic https://github.com/polkadot-js/api/blob/master/packages/rpc-provider/src/ws/index.ts#L355). Until now, that test was left out of the `integration` test directory and was never run.
> 
> 

Can you check with Nikolai? I think Jenkins allows us to trigger an action that would tag the repo when the pipeline on master succeeds automatically. Until, then we can tag manually.

If we go for the manual version, for now please add a ticket for Nikolai to add the automatic tagging to Jenkins.

.github/workflows/test-release.yml
```diff
@@ -14,4 +17,4 @@ jobs:
             - name: Run and set up the parachain, oracle, staked relayer and vault
               run: docker-compose up -d
             - run: yarn install
-            - run: yarn ci:test
+            - run: yarn ci:test:release
```
nud3l, line 20: This should run all tests on master, so just 'yarn test' I think

package.json
```diff
@@ -17,15 +17,17 @@
...
+    "ci:test:staging": "run-s build test:lint test:unit test:integration:staging",
+    "ci:test:release": "run-s build test:integration:release",
     "ci:test-with-coverage": "nyc -r lcov -e .ts -x \"*.test.ts\" yarn ci:test",
     "docs": "./generate_docs",
     "generate:defs": "ts-node --skip-project node_modules/.bin/polkadot-types-from-defs --package @interlay/polkabtc/interfaces --input ./src/interfaces",
     "generate:meta": "ts-node --skip-project node_modules/.bin/polkadot-types-from-chain --package @interlay/polkabtc/interfaces --endpoint ./src/json/parachain.json --output ./src/interfaces",
     "test": "run-s build test:*",
     "test:lint": "eslint src --ext .ts",
     "test:unit": "mocha test/unit/*.test.ts test/unit/**/*.test.ts",
-    "test:integration": "mocha test/integration/**/*.test.ts --timeout 10000000",
```
nud3l, line 28: We could leave this to run both? Maybe you'd want to do that locally in a single command

src/parachain/collateral.ts
```diff
@@ -13,25 +13,25 @@ export interface CollateralAPI {
      */
     setAccount(account: AddressOrPair): void;
     /**
-     * @returns Total locked DOT collateral
+     * @returns Total locked collateral
      */
-    totalLockedDOT(): Promise<Balance>;
+    totalLocked(): Promise<BN>;
```
nud3l, line 18: Thinking about this, would it make more sense to return a Big denominated in DOT actually? We are mostly just handing BTC and DOT amounts in the UI and there is little advantage to have manual denomination conversions

.github/workflows/test-release.yml
savudani8: Since the staging tests passed in the PR, I don't think we need to re-run them on `master`

nud3l: following scenario I have in mind:
PR1 changes staging tests X
PR2 changes staging tests Y

Both are passing. But there is an edge case where PR2 affects tests in PR1 that have just changed. Now, independently they pass, but on master they could fail. Example could be when we change the way that issue fees are deducted. In PR1, we could have fixed the issue tests. PR2 could be update of the vault struct. But maybe in the PR2, we added a new test that gets all the vault structs with the issuable tokens of the vault. But now that test would be incorrect because of PR1.

It's best practice to run all tests before a release to cover all of these edge cases.

savudani8:
The APY fix increases the displayed value by ~100x (e.g. from `0.0068` to `0.235` on my local UI)

nud3l (commented):
Pending the below: what actually changed in the calculation? Keep in mind that the UI will fetch the exchange rate globally and we could possibly just submit it as a parameter to make sure this call takes less time (i.e. on app load we already have the exchange rate so when we open the vault dashboard, we would not have to wait for this APY call to return the exchange rate promise since we don't have an API call).

It was submitted as a parameter before, but was that part incorrect since the logic itself did not change?

savudani8:
> Keep in mind that the UI will fetch the exchange rate globally and we could possibly just submit it as a parameter to make sure this call takes less time (i.e. on app load we already have the exchange rate so when we open the vault dashboard, we would not have to wait for this APY call to return the exchange rate promise since we don't have an API call).


Good point, I'll leave the `dotToBtcRate` as an optional parameter though, in case the caller does not have it (I suggested a similar approach in the requests aggregator PR, when talking about `UI Flow / Caching` https://github.com/interlay/polkabtc-js/pull/272#issuecomment-806080122). 

> Pending the below: what actually changed in the calculation? 
> It was submitted as a parameter before, but was that part incorrect since the logic itself did not change?

What changed is that previously, we were passing small denominations (sat and planck) but converting using the DOT/BTC rate, whereas now we actually pass the large denominations (hence the `Big`). 

Here is an example of where the call is wrong: https://github.com/interlay/polkabtc-js/blob/master/src/parachain/staked-relayer.ts#L385

The APY fix commit returns Big values instead of BN-like values when fetching client fees: https://github.com/interlay/polkabtc-js/pull/277/commits/d19167912c298b865b5466a9a528e52e8cd45b18#diff-63cf19a974a0d655c3d6dfbe5de1fcbe12c399d0cb8b0dce382a1c04ad30ac4dR379

nud3l:
> > Keep in mind that the UI will fetch the exchange rate globally and we could possibly just submit it as a parameter to make sure this call takes less time (i.e. on app load we already have the exchange rate so when we open the vault dashboard, we would not have to wait for this APY call to return the exchange rate promise since we don't have an API call).
> 
> Good point, I'll leave the `dotToBtcRate` as an optional parameter though, in case the caller does not have it (I suggested a similar approach in the requests aggregator PR, when talking about `UI Flow / Caching` [#272 (comment)](https://github.com/interlay/polkabtc-js/pull/272#issuecomment-806080122)).

Yep, that seems the best approach!

> 
> > Pending the below: what actually changed in the calculation?
> > It was submitted as a parameter before, but was that part incorrect since the logic itself did not change?
> 
> What changed is that previously, we were passing small denominations (sat and planck) but converting using the DOT/BTC rate, whereas now we actually pass the large denominations (hence the `Big`).
> 
> Here is an example of where the call is wrong: https://github.com/interlay/polkabtc-js/blob/master/src/parachain/staked-relayer.ts#L385
> 
> The APY fix commit returns Big values instead of BN-like values when fetching client fees: [d191679#diff-63cf19a974a0d655c3d6dfbe5de1fcbe12c399d0cb8b0dce382a1c04ad30ac4dR379](https://github.com/interlay/polkabtc-js/commit/d19167912c298b865b5466a9a528e52e8cd45b18#diff-63cf19a974a0d655c3d6dfbe5de1fcbe12c399d0cb8b0dce382a1c04ad30ac4dR379)

Ah, right makes sense! Thanks for fixing it!

nud3l (approved):

PR #278, opened 2021-04-09
feat(collateral, treasury): Subscribe to balance updates

savudani8:
Also constrains set accounts to be of type `IKeyringPair` instead of `AddressOrPair`, to remove ambiguity and ensure transactions can be signed with it. While not being very significant, this change touches many files.

I'm thinking of removing the `setAccount` pattern entirely and having the signer passed as a parameter. This would remove the is-account-set checks (e.g. here https://github.com/interlay/polkabtc-js/blob/master/src/parachain/issue.ts#L132) which are thrown if a transaction call is copy-pasted to a place where the API account isn't already set.

### DIFF
--- a/src/parachain/collateral.ts
+++ b/src/parachain/collateral.ts
@@ -1,7 +1,7 @@
 import { AccountId } from "@polkadot/types/interfaces/runtime";
 import { ApiPromise } from "@polkadot/api";
-import { ACCOUNT_NOT_SET_ERROR_MESSAGE, planckToDOT, Transaction } from "../utils";
-import { AddressOrPair } from "@polkadot/api/submittable/types";
+import { ACCOUNT_NOT_SET_ERROR_MESSAGE, dotToPlanck, planckToDOT, Transaction } from "../utils";
+import { IKeyringPair } from "@polkadot/types/types";
 import Big from "big.js";
 /**
  * @category PolkaBTC Bridge
@@ -11,7 +11,7 @@ export interface CollateralAPI {
      * Set an account to use when sending transactions from this API
      * @param account Keyring account
      */
-    setAccount(account: AddressOrPair): void;
+    setAccount(account: IKeyringPair): void;
     /**
      * @returns Total locked collateral
      */
@@ -27,17 +27,24 @@ export interface CollateralAPI {
      */
     balance(id: AccountId): Promise<Big>;
     /**
+     * Subscribe to free balance updates
+     * @param account AccountId string
+     * @param callback Function to be called whenever the balance of an account is updated.
+     * Its parameters are (accountIdString, freeBalance)
+     */
+     subscribeToBalance(account: string, callback: (account: string, balance: Big) => void): Promise<() => void>;
+    /**
      * Send a transaction that transfers from the caller's address to another address
      * @param address The recipient of the transfer
      * @param amount The balance to transfer
      */
-    transfer(address: string, amount: string | number): Promise<void>;
+    transfer(address: string, amount: Big): Promise<void>;
 }
 
 export class DefaultCollateralAPI implements CollateralAPI {
     transaction: Transaction;
 
-    constructor(private api: ApiPromise, private account?: AddressOrPair) {
+    constructor(private api: ApiPromise, private account?: IKeyringPair) {
         this.transaction = new Transaction(api);
     }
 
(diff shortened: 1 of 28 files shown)


savudani8:
I looked some more into how transaction signing is done and I believe we should keep using `AddressOrPair` instead of narrowing the type. `AddressOrPair` allows for either signing transactions with a `KeyringPair` object (which has a `sign` method because it has the private key), or specifying a string address that should sign and then doing signing via the wallet extension.

To remove duplication, however, I made all classes that send transactions inherit from TransactionAPI. This way, checking if the account is set and setting the account are only done in one place.

src/parachain/collateral.ts
```diff
@@ -26,19 +21,25 @@ export interface CollateralAPI {
...
+     * Subscribe to free balance updates
+     * @param account AccountId string
+     * @param callback Function to be called whenever the balance of an account is updated.
+     * Its parameters are (accountIdString, freeBalance)
+     */
+     subscribeToBalance(account: string, callback: (account: string, balance: Big) => void): Promise<() => void>;
     /**
      * Send a transaction that transfers from the caller's address to another address
      * @param address The recipient of the transfer
      * @param amount The balance to transfer
```
nud3l, line 34: please add "denominated in DOT"

src/parachain/collateral.ts
```diff
@@ -59,15 +60,25 @@ export class DefaultCollateralAPI implements CollateralAPI {
...
+        // as a fallback, return an empty void function
+        return () => {
+            return;
+        };
     }
 
-    setAccount(account: AddressOrPair): void {
-        this.account = account;
+    async transfer(address: string, amount: Big): Promise<void> {
+        const amountSmallDenomination = this.api.createType("Balance", dotToPlanck(amount.toString()));
```
nud3l, line 80: I guess for a future PR, we should change this to take `Big` and return a `BN`

src/parachain/collateral.ts
savudani8: Would we specify this in the description of methods or of the encompassing class?

src/parachain/collateral.ts
savudani8: Would the `BN` represent the remaining balance? In this case I think `Big` would be appropriate (it may be reused immediately to do another transfer)

src/parachain/collateral.ts
nud3l: On the method, it has the advantage is that you can see the doc comments in VSCode on the function. But it's duplication. What do you think? I think having it in the class should normally suffice and we just stick throughout the code base.

savudani8: Switching to the monetary lib will clarify expected denominations. Since that lib is close to actually being created, I think this duplication won't bring much benefit. I'll mention the denomination here, while in other places I'll just specify in the class description.

src/parachain/collateral.ts
nud3l: I was thinking `BN` just because the smallest denomination can't have any decimal places. So the transfer can on the chain side take at the bare minimum `1` (not considering dust values).

savudani8: Oh, so `transfer` would return the Planck amount that actually got submitted?
If an amount smaller than `1` is passed, a runtime error is thrown by the `Balance` constructor on this line: `REGISTRY: Error: Balance: Negative number passed to unsigned type`

nud3l: Transfer on the parachain takes the planck and it must not have decimals. I guess the random things you see is because of the weak type checking of the DOT type that somehow converts this to an integer. On our side, we should make sure that it is a BN to ensure we don't have any decimal places.

src/parachain/issue.ts
```diff
@@ -149,21 +144,26 @@ export class DefaultIssueAPI implements IssueAPI {
 
     }
 
-    async execute(issueId: H256, txId: H256Le, merkleProof: Bytes, rawTx: Bytes): Promise<void> {
-        if (!this.account) {
-            return Promise.reject(ACCOUNT_NOT_SET_ERROR_MESSAGE);
-        }
-        const executeIssueTx = this.api.tx.issue.executeIssue(issueId, txId, merkleProof, rawTx);
-        await this.transaction.sendLogged(executeIssueTx, this.account, this.api.events.issue.ExecuteIssue);
+    async execute(requestId: string, btcTxId: string): Promise<void> {
```
nud3l, line 147: Let's make the `merkleProof` and the `rawTx` optional parameter. If someone really wants to fetch their own proos, we should allow that. There are some people that might not trust blockstream with their API and would prefer fetching this directly from BTC core.

src/parachain/issue.ts
```diff
@@ -149,21 +150,22 @@ export class DefaultIssueAPI implements IssueAPI {
...
     }
 
-    async execute(issueId: H256, txId: H256Le, merkleProof: Bytes, rawTx: Bytes): Promise<void> {
-        if (!this.account) {
-            return Promise.reject(ACCOUNT_NOT_SET_ERROR_MESSAGE);
+    async execute(requestId: string, btcTxId: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void> {
+        const parsedRequestId = this.api.createType("H256", "0x" + requestId);
+        const parsedBtcTxId = this.api.createType(
+            "H256",
+            "0x" + Buffer.from(btcTxId, "hex").reverse().toString("hex")
```
nud3l, line 157: Minor nit: You can shorten this to: `Buffer.from(btcTxId, "hex").reverse()`

See UI for example https://github.com/interlay/polkabtc-ui/blob/dev/src/pages/app/issue/modal/status-view.tsx#L66

Happy to also do that in a future PR.

nud3l (approved):

src/parachain/issue.ts
savudani8: This PR is already fairly big, so I think it's best to do this in a separate PR (coupled with the `transfer`/`BN` return value)

PR #279, opened 2021-04-14
Make test utilities reusable

savudani8:
```
 rename src/external/{btc-core.ts => electrs.ts} (95%)
 rename src/{utils => parachain}/transaction.ts (91%)
 rename {test => src}/utils/bitcoin-core-client.ts (82%)
```
Also renames `BTCCoreAPI` to `electrsAPI`.
Moreover, the docker-compose dependency on `testdata-gen` is removed, to reduce nondeterminism in tests. Now the state is set via JS only.
The refund tests used to fail randomly. Now I no longer try to manually `executeIssue` and they seem to pass consistently.


src/external/electrs.ts
```diff
@@ -91,7 +91,7 @@ export interface BTCCoreAPI {
      *
      * @returns A Bitcoin transaction ID
      */
-    getTxIdByOpReturn(opReturn: string, recipientAddress?: string, amountAsBTC?: string): Promise<string>;
+    getTxIdByOpReturn(opReturn: string, recipientAddress?: string, amountAsBTC?: Big): Promise<string>;
```
nud3l, line 94: minor nit: rename to `amount` same as in the function def. below

src/external/electrs.ts
```diff
@@ -250,14 +267,30 @@ export class DefaultBTCCoreAPI implements BTCCoreAPI {
...
+            this.getTxIdByOpReturn(data)
+                .then(resolve)
+                .catch((_error) => {
+                    setTimeout(() => {
+                        console.log("Did not find opreturn, retrying...");
+                        if(timeoutMs < retryIntervalMs) {
+                            reject("Timeout elapsed");
+                        }
+                        this.waitForOpreturn(data, timeoutMs - retryIntervalMs, retryIntervalMs).then(resolve);
+                    }, 5000);
```
nud3l, line 289: shouldn't the 5000 be the `retryInterval`

docker-compose.yml
```diff
@@ -147,26 +147,3 @@ services:
                 vault --keyring=bob  --auto-register-with-collateral 1000000000000 --polka-btc-url 'ws://polkabtc:9944'
         environment:
             <<: *client-env
-    testdata_gen:
```
nud3l, line 150: Awesome to get rid of this!

src/utils/constants.ts
```diff
@@ -6,3 +6,7 @@ export const FIXEDI128_SCALING_FACTOR = 18;
 export const IGNORED_ERROR_MESSAGES = ["issue.IssueCompleted"];
 
 export const ACCOUNT_NOT_SET_ERROR_MESSAGE = "cannot request without setting account";
+
+export const mainnetApiBasePath = "https://blockstream.info/api";
```
nud3l, line 10: minor nit: should use capitalized snake case for naming

src/external/electrs.ts
gregdhill: maybe this should also be a constant?

src/utils/issue.ts
```diff
@@ -19,36 +20,27 @@ export interface IssueResult {
...
 ): Promise<IssueResult> {
     const treasuryAPI = new DefaultTreasuryAPI(api);
-    const issueAPI = new DefaultIssueAPI(api, bitcoin.networks.regtest, btcCoreAPI);
+    const issueAPI = new DefaultIssueAPI(api, bitcoin.networks.regtest, electrsAPI);
     const collateralAPI = new DefaultCollateralAPI(api);
 
-    const requester = keyring.addFromUri("//" + requesterName);
-    issueAPI.setAccount(requester);
-    const requesterAccountId = api.createType("AccountId", requester.address);
+    issueAPI.setAccount(issuingAccount);
```
gregdhill, line 35: Not sure about this API, I think the signer should only be set on init.

savudani8: It depends. If the user selects a different account from the UI, we shouldn't have to re-initialize all the APIs, but rather just call `setAccount`. Having this exposed also helps in the integration tests

gregdhill: Could the caller not call `setAccount` prior though?

gregdhill: > we shouldn't have to re-initialize all the APIs

I think the overhead is fairly minimal compared to what we would need to read from stats

gregdhill: My primary gripe is that this approach feels inconsistent if we have different signers on each component.

savudani8: This could be solved by passing an `account` reference (kind of like a singleton) to the TransactionAPI constructor (the parent class of the other APIs). In this setup, when `issueAPI.setAccount(issuingAccount);` is called, the account reference would change for all other APIs. What do you think?

savudani8: This would however count on the fact that the API objects are instantiated with the same account references - which would be easy to do in `polkabtc-api.ts` https://github.com/interlay/polkabtc-js/blob/master/src/polkabtc-api.ts#L76

test/integration/parachain/staging/refund.test.ts
```diff
@@ -30,12 +37,11 @@ describe("refund", () => {
     it("should not generate a refund request", async () => {
         const isueResult = await issue(
```
gregdhill, line 38:
```suggestion
        const issueResult = await issue(
```

PR #282, opened 2021-05-04
refactor: Add network param to issue util

savudani8:


src/utils/issue.ts
```diff
@@ -26,10 +26,26 @@ export async function issue(
     amount: Big,
     vaultAddress?: string,
     autoExecute = true,
-    triggerRefund = false
+    triggerRefund = false,
+    network?: string,
 ): Promise<IssueResult> {
     const treasuryAPI = new DefaultTreasuryAPI(api);
-    const issueAPI = new DefaultIssueAPI(api, bitcoinjs.networks.regtest, electrsAPI);
+    let bitcoinjsNetwork: bitcoinjs.Network;
```
nud3l, line 33: I think we already have this on the main API. Why do we need this here?

savudani8: I tried to use as few classes as possible. I could instantiate the entire `PolkaBTCAPI` object instead - it would be cleaner but "heavier"

nud3l: How about you have this as a util function so that instead of replicating this function (https://github.com/interlay/polkabtc-js/blob/534f3b79dfb1ba53c27fb6e54d48df06d8329393/src/polkabtc-api.ts#L24), you have a common function where this is done and then you don't need the full API object?

nud3l (approved):

PR #283, opened 2021-05-05
feat(btc-client): Mempool info

savudani8:


src/utils/bitcoin-core-client.ts
```diff
@@ -8,7 +8,7 @@ interface RecipientsToUTXOAmounts {
     [key: string]: string;
 }
 export class BitcoinCoreClient {
-    private client: typeof Client;
+    client: typeof Client;
```
theodorebugnet, line 11: Just to double-check, is exposing this intentional?

savudani8: Yep, in case I need some other quick functionalities from the bitcoin client in the future I'd like to use them directly rather than open another PR here. If those functionalities become important enough they'll make their way into a PR

theodorebugnet (approved):

PR #284, opened 2021-05-07
Parachain upgrade

savudani8:


src/parachain/btc-relay.ts
```diff
@@ -14,6 +14,11 @@ export interface BTCRelayAPI {
      * for a transaction to be considered stable
      */
     getStableBitcoinConfirmations(): Promise<number>;
+    /**
+     * @returns A global security parameter: the required block confirmations
```
nud3l, line 18: nitpick: add "on the parachain"

src/parachain/issue.ts
```diff
@@ -157,16 +159,10 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
...
         const parsedRequestId = this.api.createType("H256", "0x" + requestId);
-        const parsedBtcTxId = this.api.createType(
-            "H256",
-            "0x" + Buffer.from(btcTxId, "hex").reverse().toString("hex")
-        );
-        if (!merkleProof || !rawTx) {
-            [merkleProof, rawTx] = await this.electrsAPI.getParsedExecutionParameters(btcTxId);
-        }
-        const executeIssueTx = this.api.tx.issue.executeIssue(parsedRequestId, parsedBtcTxId, merkleProof, rawTx);
+        [merkleProof, rawTx] = await getTxProof(this.electrsAPI, btcTxId, merkleProof, rawTx);
```
nud3l, line 164: In case a user does not submit the btcTxId, I'd assume you need to fetch this here? Also this call can be optional in case the user submits the proof and rawTX

src/parachain/issue.ts
```diff
@@ -204,16 +200,15 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI
         return mapForUser;
     }
 
-    async getGriefingCollateralInPlanck(amountSat: PolkaBTC): Promise<Big> {
+    async getGriefingCollateralInPlanck(amountSat: Issuing): Promise<Big> {
```
nud3l, line 203: minor nit: use `Big` for the input as well and denominated in BTC

src/parachain/redeem.ts
```diff
@@ -52,19 +53,18 @@ export interface RedeemAPI extends TransactionAPI {
      * @param btcAddressEnc Bitcoin address where the redeemed BTC should be sent
      * @returns An object of type {redeemId, vault} if the request succeeded. The function throws an error otherwise.
      */
-    request(amount: PolkaBTC, btcAddressEnc: string, vaultId?: AccountId): Promise<RequestResult>;
+    request(amount: Issuing, btcAddressEnc: string, vaultId?: AccountId): Promise<RequestResult>;
```
nud3l, line 56: Use `Big` for the requested amount denoted in BTC

src/parachain/redeem.ts
```diff
@@ -124,7 +124,7 @@ export interface RedeemAPI extends TransactionAPI {
      * @returns The minimum amount of btc that is accepted for redeem requests; any lower values would
      * risk the bitcoin client to reject the payment
      */
-    getDustValue(): Promise<PolkaBTC>;
+    getDustValue(): Promise<Issuing>;
```
nud3l, line 127: Return `Big`

src/parachain/redeem.ts
```diff
@@ -196,16 +196,10 @@ export class DefaultRedeemAPI extends DefaultTransactionAPI implements RedeemAPI
...
         const parsedRequestId = this.api.createType("H256", "0x" + requestId);
-        const parsedBtcTxId = this.api.createType(
-            "H256",
-            "0x" + Buffer.from(btcTxId, "hex").reverse().toString("hex")
-        );
-        if (!merkleProof || !rawTx) {
-            [merkleProof, rawTx] = await this.electrsAPI.getParsedExecutionParameters(btcTxId);
-        }
-        const executeRedeemTx = this.api.tx.redeem.executeRedeem(parsedRequestId, parsedBtcTxId, merkleProof, rawTx);
+        [merkleProof, rawTx] = await getTxProof(this.electrsAPI, btcTxId, merkleProof, rawTx);
```
nud3l, line 201: same as on issue, I think you'd need to get the btxTxId in case the user does not provide it in the function call. Also this call can be optional in case the user submits the proof and rawTX

src/parachain/refund.ts
```diff
@@ -38,10 +53,24 @@ export interface RefundAPI {
...
-export class DefaultRefundAPI implements RefundAPI {
-    constructor(private api: ApiPromise, private btcNetwork: Network) { }
+export class DefaultRefundAPI extends DefaultTransactionAPI implements RefundAPI {
+    constructor(api: ApiPromise, private btcNetwork: Network, private electrsAPI: ElectrsAPI, account?: AddressOrPair) {
+        super(api, account);
+    }
+
+    async execute(requestId: string, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void> {
+        const parsedRequestId = this.api.createType("H256", "0x" + requestId);
+        [merkleProof, rawTx] = await getTxProof(this.electrsAPI, btcTxId, merkleProof, rawTx);
```
nud3l, line 70: same as on issue and redeem -> check if user submitted btxTxId and make merkleProof and rawTx getter optional

src/parachain/replace.ts
```diff
@@ -63,24 +61,43 @@ export interface ReplaceAPI extends TransactionAPI {
      * @param amountSat Amount issued, in Satoshi, to have replaced by another vault
      * @returns The request id
      */
-     request(amountSat: PolkaBTC): Promise<string>;
+    request(amountSat: Issuing): Promise<string>;
```
nud3l, line 64: `Big`

src/parachain/replace.ts
```diff
@@ -44,7 +42,7 @@ export interface ReplaceAPI extends TransactionAPI {
      * @returns The minimum amount of btc that is accepted for replace requests; any lower values would
      * risk the bitcoin client to reject the payment
      */
-    getBtcDustValue(): Promise<PolkaBTC>;
+    getBtcDustValue(): Promise<Issuing>;
```
nud3l, line 45: `Big`

src/parachain/replace.ts
```diff
@@ -63,24 +61,43 @@ export interface ReplaceAPI extends TransactionAPI {
...
      * @param account Keyring account
      */
     setAccount(account: AddressOrPair): void;
     /**
      * Wihdraw a replace request
-     * @param requestId The id of the replace request to withdraw (cancel)
+     * @param amountSat The amount of Issuing tokens to withdraw from the amount
+     * requested to have replaced.
+     */
+    withdraw(amountSat: Issuing): Promise<void>;
```
nud3l, line 75: `Big`

src/parachain/staked-relayer.ts
```diff
@@ -113,45 +89,25 @@ export interface StakedRelayerAPI extends TransactionAPI {
...
+     * A Staked Relayer reports misbehavior by a Vault, providing a fraud proof
+    * (malicious Bitcoin transaction and the corresponding transaction inclusion proof).
+     * @remarks If `txId` is not set, the `merkleProof` and `rawTx` must both be set.
+    * 
+    * @param vault_id The account of the vault to check.
+    * @param tx_id The hash of the transaction
+    * @param merkle_proof The proof of tx inclusion.
+    * @param raw_tx The raw Bitcoin transaction.
+     */
+    reportVaultTheft(vaultId: string, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void>;
```
nud3l, line 110: In this case, we need the btcTxId since I don't think there is any way for us to figure that out from just the information we have available in the lib at that point.

src/utils/bitcoin.ts
```diff
@@ -75,3 +78,13 @@ export function btcAddressFromParams(
...
     });
 }
+
+export async function getTxProof(electrsAPI: ElectrsAPI, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<[Bytes, Bytes]> {
+    if (!merkleProof || !rawTx) {
+        if(!btcTxId) {
+            throw new Error("Either the `btcTxId` or both `merkleProof` and `rawTx` must be defined to execute.");
+        }
+        [merkleProof, rawTx] = await electrsAPI.getParsedExecutionParameters(btcTxId);
+    }
```
nud3l, line 88: Scrap my comment on issue, redeem and replace, I see you are handling this here.

One thing though: we can technically also fetch everything in case the user only submits the rawTx. But it's fine if we require both I guess. Technically you can generate the btc tx id from the rawTx but it requires some extra parsing.

nud3l (approved):

PR #290, opened 2021-05-26
chore: Upgrade Big.js type dep

savudani8:


nud3l (approved):

package.json
```diff
@@ -42,7 +42,7 @@
     "@interlay/polkabtc-types": "0.7.2",
     "@polkadot/api": "4.8.1",
     "@polkadot/typegen": "4.8.1",
-    "@types/big.js": "4.0.5",
+    "@types/big.js": "6.0.3",
```
theodorebugnet, line 45: Does 6.0.3 actually exist? https://www.npmjs.com/package/@types/big.js
There's 6.0.2, and then 6.1.0. We could upgrade big.js to 6.1, but otherwise it might be safer to use types 6.0.2 for now

PR #292, opened 2021-05-27
fix(replace): Griefing collateral scaling

savudani8:


src/parachain/replace.ts
```diff
@@ -120,8 +120,9 @@ export class DefaultReplaceAPI extends DefaultTransactionAPI implements ReplaceA
 
     async request(amount: Big): Promise<string> {
         const amountSat = this.api.createType("Wrapped", btcToSat(amount.toString()));
-        const griefingCollateral = await this.getGriefingCollateral(amount);
-        const requestTx = this.api.tx.replace.requestReplace(amountSat, btcToSat(griefingCollateral.toString()));
+        const griefingCollateralDot = await this.getGriefingCollateral(amount);
+        const griefingCollateralPlanck = dotToPlanck(griefingCollateralDot.toString()) as string;
```
gregdhill, line 124: Not a problem for today, but this is nasty: `toString`, `as string`.

gregdhill (approved):

src/parachain/replace.ts
savudani8: `dotToPlanck` returns `undefined` in some cases, but if that were the case here an error should already have been thrown

savudani8: Oh, you mean converting to/from strings all the time. I agree, Dom already pointed out we should use a number type. I'll add a ticket

nud3l (approved):

PR #294, opened 2021-05-27
feat: add active block number

nud3l:


src/parachain/system.ts
```diff
@@ -8,14 +8,23 @@ export interface SystemAPI {
...
-    async getCurrentBlockNumber(): Promise<BlockNumber> {
+    async getCurrentBlockNumber(): Promise<number> {
+        const head = await this.api.rpc.chain.getFinalizedHead();
+        return await (await this.api.query.system.number.at(head)).toNumber();
+    }
+
+    async getCurrentActiveBlockNumber(): Promise<number> {
         const head = await this.api.rpc.chain.getFinalizedHead();
-        return await this.api.query.system.number.at(head);
+        return await (await this.api.query.security.activeBlockCount.at(head)).toNumber();
```
savudani8, line 28: The first `await` is not required, but it doesn't break anything

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/polkabtc",
-  "version": "0.15.6",
+  "version": "0.16.0",
```
savudani8, line 3: 0.15.7 will soon be merged, hope it won't cause a conflict

src/parachain/system.ts
nud3l: uh, that was the auto-complete, I'll fix that

package.json
nud3l: yeah, no worries I'll fix that

PR #298, opened 2021-05-28
fix: get issuable amount for specific vault

gregdhill:

### DIFF
--- a/src/parachain/vaults.ts
+++ b/src/parachain/vaults.ts
@@ -148,14 +148,19 @@ export interface VaultsAPI extends TransactionAPI {
      */
     getIssuedAmount(vaultId: AccountId): Promise<Big>;
     /**
+     * @param vaultId The vault account ID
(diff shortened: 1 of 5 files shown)


src/parachain/vaults.ts
```diff
@@ -435,14 +434,27 @@ export class DefaultVaultsAPI extends DefaultTransactionAPI implements VaultsAPI
         return new Big(satToBTC(vault.issued_tokens.toString()));
     }
 
-    private async getIssuedPolkaBTCAmounts(): Promise<Big[]> {
+    async getIssuableAmount(vaultId: AccountId): Promise<Big> {
```
savudani8, line 437: This function should be added to the interface so the UI can access it

gregdhill: I think it already is right? Or is there another interface I missed?

savudani8 (approved):

nud3l (approved):

PR #300, opened 2021-06-01
chore(test): Upgrade clients

savudani8:
Also extends redeem subscriptions to add `subscribeToRedeemCompletion` (not yet exposed in the interface, needs further testing)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.8"
 services:
     polkabtc:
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.8"
 services:
     polkabtc:
-        image: "registry.gitlab.com/interlay/btc-parachain:0-7-4-7c02930d-1622122540"
+        image: "registry.gitlab.com/interlay/btc-parachain:master-c26d3cc5-1622628942"
```
gregdhill, line 4: We can / should use `0.7.5` for the parachain.

PR #304, opened 2021-07-06
added API to subscribe to new finalized block heads with a callback

theodorebugnet:
Stats/Index needs to subscribe to the latest parachain block number. I exposed a subscription to the entire header in case it becomes useful elsewhere.

Additionally, adds `block_height` return to `getTransactionStatus()`, to optimise away the extra call to `getTransactionBlockHeight` in cases where both the status and the block height are necessary.

### DIFF
--- a/src/parachain/system.ts
+++ b/src/parachain/system.ts
@@ -1,4 +1,5 @@
 import { ApiPromise } from "@polkadot/api";
+import { Header } from "@polkadot/types/interfaces";
 
 /**
  * @category InterBTC Bridge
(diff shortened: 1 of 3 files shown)


src/parachain/system.ts
```diff
@@ -12,6 +13,12 @@ export interface SystemAPI {
      * @returns The current active block number being processed.
      */
     getCurrentActiveBlockNumber(): Promise<number>;
+
+    /**
+     * On every new parachain block, call the callback function with the ID of the expired request.
```
savudani8, line 18: This comment should be changed as it's outdated

theodorebugnet: What should it be changed to? As far as I understand it, the subscription does exactly that, call the callback on every new finalised header. Does that not happen on every block?

> it's outdated

I just wrote it about an hour ago, so it could easily be completely wrong but I don't think it's "outdated" haha

savudani8:
```suggestion
     * On every new parachain block, call the callback function with the new block header
```
Sorry about the wording - by outdated I meant that it might have been copied over from another TS-Doc comment

theodorebugnet: Ooh, right. You're absolutely correct, thanks.

savudani8 (approved):

PR #305, opened 2021-07-07
Monetary.js integration

savudani8:
- Of particular note is how the Tokens module abstracts multiple currencies. This pattern is reused across several files.
- There are also some changes in `src/types/currency.ts` to accommodate the abstraction.
- This PR attempts to add most of the logic required for multi-collateral

### DIFF
--- a/src/external/electrs.ts
+++ b/src/external/electrs.ts
@@ -13,11 +13,9 @@ import { AxiosResponse } from "axios";
 import * as bitcoinjs from "bitcoinjs-lib";
 import { TypeRegistry } from "@polkadot/types";
 import { Bytes } from "@polkadot/types";
-import Big from "big.js";
-import BN from "bn.js";
+import { BTCAmount } from "@interlay/monetary-js";
 
 import { MAINNET_ESPLORA_BASE_PATH, REGTEST_ESPLORA_BASE_PATH, TESTNET_ESPLORA_BASE_PATH } from "../utils/constants";
-import { btcToSat } from "../utils/currency";
 
 export type TxStatus = {
     confirmed: boolean;
@@ -94,7 +92,7 @@ export interface ElectrsAPI {
      *
      * @returns A Bitcoin transaction ID
      */
-    getTxIdByOpReturn(opReturn: string, recipientAddress?: string, amount?: Big): Promise<string>;
+    getTxIdByOpReturn(opReturn: string, recipientAddress?: string, amount?: BTCAmount): Promise<string>;
     /**
      * Fetch the last bitcoin transaction ID based on the recipient address and amount.
      * Throw an error if no such transaction is found.
@@ -103,11 +101,11 @@ export interface ElectrsAPI {
      * Performs the lookup using an external service, Esplora
      *
      * @param recipientAddress Match the receiving address of a UTXO
-     * @param amountAsBTC Match the amount (in BTC) of a UTXO that contains said recipientAddress.
+     * @param amount Match the amount (in BTC) of a UTXO that contains said recipientAddress.
      *
      * @returns A Bitcoin transaction ID
      */
-    getTxIdByRecipientAddress(recipientAddress: string, amountAsBTC?: Big): Promise<string>;
+    getTxIdByRecipientAddress(recipientAddress: string, amount?: BTCAmount): Promise<string>;
     /**
      * Fetch the Bitcoin transaction that matches the given TxId
      *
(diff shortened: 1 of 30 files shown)


src/external/electrs.ts
```diff
@@ -240,17 +238,19 @@ export class DefaultElectrsAPI implements ElectrsAPI {
...
-    private utxoHasAtLeastAmount(utxo: UTXO | VOut, amount?: Big): boolean {
+    private utxoHasAtLeastAmount(utxo: UTXO | VOut, amount?: BTCAmount): boolean {
         if (amount) {
-            const expectedBtcAsSatoshi = btcToSat(amount);
-            if (utxo.value === undefined || expectedBtcAsSatoshi > new BN(utxo.value)) {
+            if (utxo.value === undefined) {
                 return false;
+            } else {
+                const utxoValue = BTCAmount.from.Satoshi(utxo.value);
+                return utxoValue >= amount;
```
theodorebugnet, line 247: Are you sure this will work? I'd expect one would need to use `utxoValue.gte(amount)`

src/types/currency.ts
```diff
@@ -1,12 +1,86 @@
...
+export function currencyIdToMonetary(currencyId: CurrencyIdLiteral): CollateralCurrency {
+    switch (currencyId) {
+        case CurrencyIdLiteral.DOT: {
+            return Polkadot;
+        }
+        case CurrencyIdLiteral.KSM: {
+            return Kusama;
+        }
+    }
+    throw new Error("No CurrencyId entry for provided Monetary");
```
theodorebugnet, line 56: Other way around

theodorebugnet (commented):
I'm wondering, would it make sense to define an `interBitcoin` currency, separate from `Bitcoin`? I'm not sure if it ever makes sense to e.g. add unwrapped bitcoin amounts with wrapped amounts. Or compare equality between unwrapped and wrapped BTC.

In many cases BTC != interBTC, in fact: for example, due to fees, `issueRequest.btcAmount !== issueRequest.interBTCAmount`. I'm wondering if it would make sense to add this distinction at the type level or not.

savudani8:
> In many cases BTC != interBTC, in fact: for example, due to fees, issueRequest.btcAmount !== issueRequest.interBTCAmount. I'm wondering if it would make sense to add this distinction at the type level or not.

I agree, though how easy would it be to convert between the two? Would we need to define an exchange rate of `1`?
For instance, there is this test
https://github.com/interlay/interbtc-js/blob/dan/monetary/test/integration/parachain/staging/issue.test.ts#L96

theodorebugnet:
> Would we need to define an exchange rate of 1?

Sounds about right. The tradeoff is then basically that you have to explicitly convert between the two, and otherwise they're considered separate types, so it prevents accidentally mixing the two units. But there's more overhead when you do want to compare them.

src/external/electrs.ts
savudani8: I assumed that defining `gte`, `lte`, etc. members would override the default functionality of comparison operators. But this actually compares their string representations, which is wrong, thanks for pointing it out

savudani8:
@theodorebugnet I had a look at the code and introducing the `InterBTC` currency would add enough changes to prompt a separate PR. For instance, when computing APY, we need to convert all rewards to the same currency and divide by collateral. To do the conversion, we're currently passing the BTC/DOT exchange rate as a parameter, but the reward is in interBTC. So we'd first need to convert InterBTC to BTC and then to DOT. Maybe we could determine exchange rates transitively under the hood given a chain for exchange rate objects.

theodorebugnet (approved):

theodorebugnet:
Adding automatic transitive conversions would definitely be helpful, gonna add a ticket for this.

src/parachain/btc-relay.ts
```diff
@@ -35,7 +35,7 @@ export interface BTCRelayAPI {
      * @param confirmations The number of block confirmations needed to accept the inclusion proof.
```
nud3l, line 35: Slightly update the doc comment here since we don't have the boolean anymore.

src/parachain/fee.ts
```diff
@@ -1,33 +1,48 @@
...
 import { ApiPromise } from "@polkadot/api";
+import {
+    Bitcoin,
+    BTCAmount,
+    BTCUnit,
+    Currency,
+    ExchangeRate,
+    MonetaryAmount,
+    Polkadot,
+    PolkadotAmount,
```
nud3l, line 12: Probably requires a change in the monetary lib, but the `BTCAmount` and `DOTAmount` should have consistent naming. So rather `DOTAmount` than `PolkadotAmount`. Same for the units `PolkadotUnit` -> `DOTUnit`

savudani8: Added a ticket

PR #306, opened 2021-07-07
refactor: reduce number of calls and parallelize to get system capacity

nud3l:
Refactored:
- The current way to get the system capacity would get the issued tokens for every vault. However, this is the same as the total supply of interBTC. Therefore, it's easier to just get the total of interBTC.
- The backing collateral in DOT is already contained in the vault struct, removed that extra call to get the vault twice from the chain.
- Executed some calls in parallel instead of sequentially.

### DIFF
--- a/src/parachain/vaults.ts
+++ b/src/parachain/vaults.ts
@@ -286,7 +286,7 @@ export interface VaultsAPI extends TransactionAPI {
     computeReward(currencyId: CurrencyIdLiteral, localPoolId: string, accountId: string): Promise<BN>;
     /**
      * @param vaultId account id
-     * @returns The entire collateral backing a vault's issued tokens. 
+     * @returns The entire collateral backing a vault's issued tokens.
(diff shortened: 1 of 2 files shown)


src/parachain/vaults.ts
```diff
@@ -544,7 +544,7 @@ export class DefaultVaultsAPI extends DefaultTransactionAPI implements VaultsAPI
 
     async getIssuableAmount(vaultId: AccountId): Promise<Big> {
         const vault = await this.get(vaultId);
-        const lockedDot = await this.getBackingCollateral(vaultId);
+        const lockedDot = planckToDOT(vault.backing_collateral.toBn());
```
savudani8, line 547: This was indeed a bug

savudani8:
The FIXME notes are already addressed here: https://github.com/interlay/interbtc-js/pull/305
Might me worth removing them just to minimize merge conflicts

savudani8 (approved):

PR #308, opened 2021-07-15
Expose interbtc index

theodorebugnet:
Added module exposing interbtc-index, allowing for explicit wrappers as desired and auto-generating thin wrappers for every other API in index.

Due to the necessity to create properties according to the imported client, a class couldn't be used as with the other APIs, hence this module returns a simple object.


src/utils/constants.ts
```diff
@@ -10,3 +10,5 @@ export const ACCOUNT_NOT_SET_ERROR_MESSAGE = "cannot request without setting acc
 export const MAINNET_ESPLORA_BASE_PATH = "https://blockstream.info/api";
 export const TESTNET_ESPLORA_BASE_PATH = "https://electr-testnet.do.polkabtc.io";
 export const REGTEST_ESPLORA_BASE_PATH = "http://0.0.0.0:3002";
+
+export const INDEX_URL = process.env.INTERBTC_INDEX_SERVER_URL || "http://localhost:3007";
```
savudani8, line 14: The lib is not really a standalone program, so there should be no environment variables

theodorebugnet: Good point. I'll make it a parameter.

src/external/interbtc-index.ts
```diff
@@ -0,0 +1,53 @@
...
+const GeneratedMiddlewareFns = ["withMiddleware", "withPreMiddleware", "withPostMiddleware"] as const;
+type GeneratedMiddlewareFns = typeof GeneratedMiddlewareFns[number];
+// For every 'foo() => Promise<T>' function, the generated client
+// has a 'fooRaw() => <Promise<ApiResponse<T>>' counterpart. These must be filtered out
+type RawApiResponse = Promise<interbtcIndex.ApiResponse<any>>;
+
+export type IndexAPI = Pick<
+    interbtcIndex.IndexApi,
+    {
+        [ApiFn in keyof interbtcIndex.IndexApi]-?: ReturnType<interbtcIndex.IndexApi[ApiFn]> extends RawApiResponse
```
savudani8, line 13: Just for my understanding: Just using the `Required<T>` type here would not have been enough, because what you're filtering out with `-?` are the `never`s generated in the lines below?

theodorebugnet: Yep, correct

savudani8 (approved):

PR #312, opened 2021-07-28
feat(request-timestamps): potential timestamp implementation for issue?

theodorebugnet:
Will query the parachain twice extra for every issue request!

### DIFF
--- a/src/parachain/issue.ts
+++ b/src/parachain/issue.ts
@@ -28,9 +28,16 @@ import { CollateralUnit, Issue, IssueStatus } from "../types";
 
 export type IssueLimits = { singleVaultMaxIssuable: BTCAmount; totalMaxIssuable: BTCAmount };
 
-export function encodeIssueRequest(req: IssueRequest, network: Network, id: H256 | string): Issue {
+export async function encodeIssueRequest(
+    req: IssueRequest,
+    network: Network,
+    id: H256 | string,
+    api: ApiPromise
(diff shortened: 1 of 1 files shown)


savudani8 (approved):
Up to @nud3l but since the UI will be querying `index` I think it's fine to take the performance hit

theodorebugnet:
I have the opposite opinion, since the UI will be querying index we don't really need the timestamp here, but the lib should still be usable for other people who want to integrate with the parachain. Not every single usecase will need the timestamp, and a mandatory 3x performance hit seems excessive for just one non-critical field.

savudani8:
In this case you can expose a `getTimestampFromBlocknumber()` function using the logic in this PR to let users decide whether/where to call it. A good place to add it could be `src/utils/encoding.ts`

nud3l:
I am in favor of having the timestamp on `issue.request()` but against having it on other calls for the performance penalty reason. However, with a different implementation approach:

If you were to do this accurately, you would go the way of: `request` is using `sendLogged` under the hood. This already awaits the transactions being included and includes the block hash at which the transaction is included. So instead of making another extra call to translate the block height to the hash, you can get away with just a single timestamp call.

But, that will be more or less the same as if you'd just use `Date.now()` on the request API. So I think we should just go with that.

On any list calls, I would not add it automatically due to performance issues. I like Dan's suggestions though to add this API call for getting a timestamp for a block number.

What do you think?

PR #313, opened 2021-07-30
feat(setup): Expose initialization script

savudani8:


src/utils/setup.ts
```diff
@@ -0,0 +1,266 @@
...
+const { hideBin } = require("yargs/helpers");
+const argv = yargs(hideBin(process.argv))
+    .option("initialize", {
+        type: "boolean",
+        description: "Flag that decides whether this script should be run or not",
+        default: false,
+    })
+    .option("set-stable-confirmations", {
+        type: "boolean",
+        description: "Set the required bitcoin and parachain confirmations to zero",
```
gregdhill, line 44: Description here is misleading, if I'm correct the caller can set any parameter they want?

test/config.ts
```diff
@@ -1,8 +0,0 @@
-export const DEFAULT_PARACHAIN_ENDPOINT = "ws://127.0.0.1:9944";
```
gregdhill, line 1: I think we should keep this file and just import into the setup script.

src/utils/setup.ts
gregdhill: I actually think this is not worth including if we can query from the parachain - i.e. on dev it will be zero and there is no need to set this param otherwise.

savudani8: > Description here is misleading
The CLI can't set number of confirmations, only whether to set the to zero

> on dev it will be zero and there is no need to set this param otherwise.
It's not zero on local in my experience, the tests fail if this isn't run

gregdhill: > It's not zero on local in my experience, the tests fail if this isn't run

It's probably one, but this shouldn't make a difference. If the tests fail because of this we should find out why.

savudani8: It's 1 bitcoin confirmation but 100 parachain confirmations

gregdhill: Ah that's right, perhaps we should anyway set both to zero for dev in genesis.

theodorebugnet (approved):
LGTM

PR #314, opened 2021-08-03
Typed structs

savudani8:
Depends on https://github.com/interlay/interbtc-js/pull/313


src/external/interbtc-index.ts
```diff
@@ -23,7 +24,16 @@ export const DefaultIndexAPI: (configuration: interbtcIndex.Configuration) => In
...
+        // "getVaults": () => {
+        //     index.getVaults()
+        // }
     };
+    // getVaults(
+    //     @Query() page = 0,
+    //     @Query() perPage = defaultPerPage,
+    //     @Query() sortBy: VaultColumns = "block_number",
+    //     @Query() sortAsc = false
+    // )
```
gregdhill, line 36: I think this can be removed

src/parachain/redeem.ts
```diff
@@ -42,10 +41,10 @@ export function encodeRedeemRequest(req: RedeemRequest, network: Network, id: H2
     return {
         id: stripHexPrefix(id.toString()),
         userDOTAddress: req.redeemer.toString(),
-        amountBTC: satToBTC(req.amount_btc).toString(),
-        dotPremium: planckToDOT(req.premium).toString(),
-        bridgeFee: satToBTC(req.fee).toString(),
-        btcTransferFee: satToBTC(req.transfer_fee_btc).toString(),
+        amountBTC: BTCAmount.from.Satoshi(req.amount_btc.toString()).str.BTC(),
```
gregdhill, line 44: What is `.str`?

src/types/vault.ts
```diff
@@ -0,0 +1,46 @@
...
+import { AccountId } from "@polkadot/types/interfaces";
+import * as interbtcIndex from "@interlay/interbtc-index-client";
+
+import { VaultStatus } from "../interfaces";
+import Big from "big.js";
+
+export interface WalletExt {
+    // network encoded btc addresses
+    publicKey: string;
+    btcAddress?: string;
```
gregdhill, line 11: I think this can be removed, most requests will specify the `btcAddress` for payments.

src/utils/encoding.ts
```diff
@@ -106,12 +120,92 @@ export interface DecodedRequestExt extends Omit<DecodedRequest, "btc_address"> {
...
+
+//         ...vaultData,
+//         collateral: PolkadotAmount.from.
+//         lockedBTC: BTCAmount;
+//         pendingBTC: BTCAmount;
+//         collateralization: Big;
+//         pendingCollateralization: Big;
+//         capacity: BTCAmount;
+//     }
+// }
```
gregdhill, line 210: Remove and add when required

test/integration/external/staging/electrs.test.ts
```diff
@@ -9,7 +9,8 @@ import { BitcoinCoreClient } from "../../../../src/utils/bitcoin-core-client";
 import { REGTEST_ESPLORA_BASE_PATH } from "../../../../src";
 import { BTCAmount } from "@interlay/monetary-js";
 
-describe("ElectrsAPI testnet", function () {
+// TODO: Unskip tests when the electrs instance is fixed
+describe.skip("ElectrsAPI testnet", function () {
```
gregdhill, line 13: The instance should be working now (cc: @ns212)

gregdhill (commented):
Please also resolve the compiler warnings (mostly just unused imports).

src/parachain/redeem.ts
savudani8: It's a field of Monetary.js objects that has methods for displaying the amount in all denominations (e.g. for Bitcoin, `.str.BTC()` and `.str.Satoshi()`)

test/integration/external/staging/electrs.test.ts
savudani8: I think we should actually remove the testnet tests. They only test the testnet deployment, since the same logic is also tested on the Regtest Electrs instance. This E2E functionality should be checked in DataDog or the Bot instead.

gregdhill (approved):

PR #315, opened 2021-08-09
fix: Nomination rewards; remove non-interbtc fee getters

savudani8:
The bug was on the lib side. Rewards in the `rewards` (global) pool were not scaled down by the nominator's proportion of collateral.

The PR also removes DOT rewards from APY calculation, since these will always be zero.

### DIFF
--- a/src/parachain/nomination.ts
+++ b/src/parachain/nomination.ts
@@ -2,7 +2,7 @@ import { Network } from "bitcoinjs-lib";
 import { ApiPromise } from "@polkadot/api";
 import { AddressOrPair } from "@polkadot/api/submittable/types";
 import { AccountId, Index } from "@polkadot/types/interfaces";
-import { Bitcoin, Currency, BTCAmount, MonetaryAmount } from "@interlay/monetary-js";
+import { Bitcoin, Currency, BTCAmount, MonetaryAmount, Polkadot } from "@interlay/monetary-js";
 
 import { CurrencyId, UnsignedFixedPoint } from "../interfaces";
 import { DefaultVaultsAPI, VaultsAPI } from "./vaults";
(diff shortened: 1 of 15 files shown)


theodorebugnet (approved):
I'm not super familiar with the pool APIs but the code LGTM

src/parachain/nomination.ts
```diff
@@ -102,12 +102,24 @@ export interface NominationAPI extends TransactionAPI {
      *
      * @param nominatorId Id of user who nominated to one or more vaults
      * @param currency The currency of the nominations
-     * @returns The rewards (in InterBTC a nominator has accumulated)
+     * @returns The rewards a currently active nominator has accumulated
      */
-    getNominatorRewards<C extends CurrencyUnit>(
+    getActiveNominatorRewards<C extends CurrencyUnit>(
```
gregdhill, line 107: Minor nit, old naming was fine - Nominators are never not active.

src/parachain/pools.ts
```diff
@@ -241,24 +239,43 @@ export class DefaultPoolsAPI implements PoolsAPI {
...
+        const nominatorCollateral = await this.computeCollateralInStakingPool(currency, nomineeId, nominatorId);
+        return nominatorCollateral.toBig().div(backingCollateral.toBig());
     }
 
-    async getFeesWrapped(vaultId: string): Promise<BTCAmount> {
-        return await this.computeReward(Bitcoin, vaultId, vaultId);
+    async computeReward<C1 extends CurrencyUnit, C2 extends CollateralUnit>(
+        currency: Currency<C1>,
+        collateralCurrency: Currency<C2>,
+        nomineeId: string,
```
gregdhill, line 260:
```suggestion
        vaultId: string,
```

src/parachain/pools.ts
```diff
@@ -241,24 +239,43 @@ export class DefaultPoolsAPI implements PoolsAPI {
         return decodeFixedPointType(await this.api.query.rewards.rewardPerToken.at(head, currencyId));
     }
 
-    async computeReward<C extends CurrencyUnit>(
-        currency: Currency<C>,
+    async backingCollateralProportion<C1 extends CurrencyUnit, C2 extends CollateralUnit>(
+        currency: Currency<C1>,
+        collateralCurrency: Currency<C2>,
         nomineeId: string,
```
gregdhill, line 245:
```suggestion
        vaultId: string,
```

src/parachain/nomination.ts
savudani8: They can be not active if they don't have active nomination but still have rewards to collect. Maybe another name would describe the situation better, but we need to make the distinction

gregdhill: What are we making the distinction on? Is there a `getUnactiveNominatorRewards`?

savudani8: Yes, but the naming (perhaps not the best) is `getNominatorReward` https://github.com/interlay/interbtc-js/pull/315/files#diff-43eba5e3e371acf89540a6f85d6a52f42e005c8609363a860b60be839c69d2c6R118

gregdhill: Ah I missed that, okay then makes sense to me!

gregdhill (approved):

PR #316, opened 2021-08-09
fix(package.json): fixed build paths

theodorebugnet:
Importing the lib fails for me without this (please double-check this change isn't only on my end - but I'm very consistently getting output in `build/src/*` in the latest lib version).


savudani8:
The `main` entry point should stay as it is, I think the fix should be something else. I'll check in more detail tomorrow

savudani8 (approved):

PR #317, opened 2021-08-10
refactor!: remove staked relayer sub-system

gregdhill:


src/parachain/vaults.ts
```diff
@@ -262,6 +264,17 @@ export interface VaultsAPI extends TransactionAPI {
         vaultId: AccountId,
         currency: Currency<C>
     ): Promise<MonetaryAmount<Currency<C>, C>>;
+    /**
+     * A relayer may report Vault misbehavior by providing a fraud proof
```
savudani8, line 268: Do we keep the `relayer` terminology or should we say `A vault may report another vault's misbehavior...`

gregdhill: I think `relayer` is fine since it can mean anyone who relays the theft proof, whereas `vault` would be inaccurate because anyone can submit this.

savudani8 (approved):

PR #320, opened 2021-08-10
feat: Parse issue and redeem types

savudani8:
Next step is to stringify these types in an `index` PR

### DIFF
--- a/src/utils/encoding.ts
+++ b/src/utils/encoding.ts
@@ -9,7 +9,7 @@ import { BTCAmount, PolkadotAmount } from "@interlay/monetary-js";
 
 import { encodeBtcAddress, FIXEDI128_SCALING_FACTOR } from ".";
 import { WalletExt, SystemVaultExt } from "../types/vault";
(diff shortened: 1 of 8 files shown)


src/utils/encoding.ts
```diff
@@ -171,3 +173,45 @@ export function parseReplaceRequest(req: ReplaceRequest, network: Network): Repl
...
+        griefingCollateral: PolkadotAmount.from.Planck(req.griefing_collateral.toString()),
+        status,
+    };
+}
+
+export function parseRedeemRequest(req: RedeemRequest, network: Network, id: H256): Redeem {
+    const status = req.status.isCompleted
+        ? RedeemStatus.Completed
+        : req.status.isRetried
+            ? RedeemStatus.Retried
```
theodorebugnet, line 201: Huh, I tried to use prettier to change the formatting but then eslint complained. Maybe we should look into fixing the linting configs so they can work together (later, of course).

savudani8: I completely stopped using prettier. To format, I now run `yarn fix` which applies eslint at the end.

PR #321, opened 2021-08-10
Theo/integrate index wrappers

theodorebugnet:
*Do not merge until #320 (lib version 0.23) is merged+tagged*

Added wrappers to index oracle + vaults cache
 + added local VaultStatus type, for easier construction
 + renamed awkward function in `parachain/system`
 + added minor optimisation for index use, allowing supplying block hashes to certain query functions instead of fetching the finalized head every time


savudani8 (approved):
So no version bump in this one?

theodorebugnet:
Nope, we merge yours and mine and then tag 0.23, from what I understand. Alternatively I can bump to 0.24 as well, I'm just not sure which is best.

Actually I think on second thought I'll tag 0.24, that way e.g. the UI can be updated incrementally if desired.

PR #324, opened 2021-08-17
fix(import): avoid exporting the setup script

theodorebugnet:
to prevent issues when the library is used as a module rather than an executable

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -29,9 +29,9 @@ export function createPolkadotAPI(endpoint: string, autoConnect?: number | false
 export async function createInterbtcAPI(
     endpoint: string,
     network: BitcoinNetwork = "mainnet",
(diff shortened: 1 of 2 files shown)


nud3l:
I think the idea was that we could use that to setup the docker-compose in the UI using the same scripts. What issues does this cause?

theodorebugnet:
I believe setting up the environment involves calling the lib as an independent command rather than importing it programmatically as a module. I don't think we'd ever need to call into the setup script from inside the UI code.

Re: infrastructure repo: I think this idea has been brought up a few times before, and every time everyone agrees it'd be a good idea haha. Only potential tricky part I can see is that the lib uses a different setup for its tests compared to the UI, so we'd either unify them or have two separate setups in that new repo.

PR #329, opened 2021-09-02
feat: Upgrade to parachain v1.0

savudani8:
First commit:
- Upgrade to 0.12.x
- Finishes implementation of multi-collateral 
- Adds support for multiple wrapped tokens per vault
- Where this cannot be inferred, the caller of a function needs to specify the collateral currency / wrapped currency
- The PR is big enough as it is, I didn't want to delay it by upgrading to 1.x

Second commit:
- Upgrade to 1.0
- Tests multi-collateral setup (dave_stash is backed by KSM)
- async calls have been optimized
- Usage of `BTCAmount` has been entirely refactored -> the interfaces support all wrapped BTC types
- `wrappedCollateral` parameter has been moved from the methods to the api constructor in most cases
- Updates readme

TODO in a separate PR:
- reproduce the removed `vaultRegistry.getTotalCollateralization` in the lib using storage values
- in the `vaultsAPI`, remove `collateralCurrency` parameter from methods and infer based on the set account


package.json
```diff
@@ -40,12 +40,12 @@
   "engineStrict": true,
   "dependencies": {
     "@interlay/esplora-btc-api": "0.4.0",
-    "@interlay/interbtc-types": "0.10.2",
-    "@interlay/monetary-js": "0.2.2",
+    "@interlay/interbtc-types": "0.12.3",
+    "@interlay/monetary-js": "0.2.4",
```
nud3l, line 44: Could you upgrade to the monetary lib version that already has the fix for the `DOTAmount` in it to make the naming across currencies consistent please?

src/parachain/nomination.ts
```diff
@@ -74,54 +77,53 @@ export interface NominationAPI extends TransactionAPI {
...
-        currency: Currency<C>,
+    getFilteredNominations(
         nominatorId?: string,
         vaultId?: string
-    ): Promise<[[string, string], MonetaryAmount<Currency<C>, C>][]>;
+    ): Promise<[[string, string], MonetaryAmount<Currency<CollateralUnit>, CollateralUnit>, WrappedCurrency][]>;
     /**
      * @remarks At least one of the parameters must be specified
-     * @param currency The collateral currency of the nominations
+     * @param wrappedCurrency The currency towards whose issuance the nomination was made
```
nud3l, line 90: I am wondering if it would make more sense to add the wrapped currency as a default on instantiation of the lib. Since you'd use the lib with exactly one wrapped token (i.e., either kBTC or interBTC), I think it would make sense. Also in the case where one chain would serve different wrapped assets like BCH, ZEC, I think unlikely that you were to change the wrapped asset individually per call, but rather have separate instances of the lib for each wrapped token.

src/parachain/nomination.ts
```diff
@@ -188,22 +190,36 @@ export class DefaultNominationAPI extends DefaultTransactionAPI implements Nomin
...
-    async listNominatorsRaw<C extends CurrencyUnit>(
-        currency: Currency<C>
+    async listAllNominations(): Promise<
+        [[WrappedCurrency, [number, string, string]], MonetaryAmount<Currency<CollateralUnit>, CollateralUnit>][]
+        > {
+        let result: [
+            [WrappedCurrency, [number, string, string]],
+            MonetaryAmount<Currency<CollateralUnit>, CollateralUnit>
+        ][] = [];
+        for (const currency of WRAPPED_CURRENCIES) {
```
nud3l, line 200: Yeah, actually I prefer having the wrapped currency being set on the global API level rather than returning all the nominations combined for all wrapped currencies here.

nud3l, line 200: Also, slight implementation note, instead of using a for loop here, better to use a `map` function that can execute the await call in line 201 in parallel otherwise you will execute it sequentially.

src/parachain/nomination.ts
```diff
@@ -188,22 +190,36 @@ export class DefaultNominationAPI extends DefaultTransactionAPI implements Nomin
...
+
+    async listAllNominationsForWrappedCurrency(
+        wrappedCurrency: WrappedCurrency
     ): Promise<
-        [[Currency<CurrencyUnit>, [number, string, string]], MonetaryAmount<Currency<CurrencyUnit>, CurrencyUnit>][]
+        [[WrappedCurrency, [number, string, string]], MonetaryAmount<Currency<CollateralUnit>, CollateralUnit>][]
     > {
         const head = await this.api.rpc.chain.getFinalizedHead();
-        const nonces = await this.getNonces(currency);
+        const nonces = await this.getNonces(wrappedCurrency);
```
nud3l, line 213: This call and the one above to get the head can be wrapped in a `Promise.all`

src/parachain/pools.ts
```diff
@@ -13,46 +13,47 @@ import {
     newMonetaryAmount,
     tickerToCurrencyIdLiteral,
 } from "../";
+import { WrappedCurrency } from "../types";
 import { newAccountId } from "../utils";
 
 export interface PoolsAPI {
```
nud3l, line 19: I think this should be called `RewardsAPI` to account for the staking/rewards pallet name.

src/parachain/pools.ts
```diff
@@ -92,123 +93,121 @@ export interface PoolsAPI {
...
+        vaultId: string,
         nominatorId: string
-    ): Promise<MonetaryAmount<Currency<C>, C>> {
-        const currencyId = tickerToCurrencyIdLiteral(currency.ticker);
-        const stake = await this.getStakingPoolStake(currencyId, nomineeId, nominatorId);
-        const rewardPerToken = await this.getStakingPoolRewardPerToken(currencyId, nomineeId);
-        const rewardTally = await this.getStakingPoolRewardTally(currencyId, nomineeId, nominatorId);
+    ): Promise<MonetaryAmount<Currency<BTCUnit>, BTCUnit>> {
+        const currencyId = tickerToCurrencyIdLiteral(wrappedCurrency.ticker);
+        const stake = await this.getStakingPoolStake(currencyId, vaultId, nominatorId);
```
nud3l, line 123: Wrap the calls below in a `Promise.all` please

src/parachain/pools.ts
```diff
@@ -239,43 +238,34 @@ export class DefaultPoolsAPI implements PoolsAPI {
...
-    ): Promise<MonetaryAmount<Currency<C1>, C1>> {
-        const totalGlobalReward = await this.computeRewardInRewardsPool(currency, nomineeId);
-        const globalRewardShare = await this.backingCollateralProportion(
-            currency,
-            collateralCurrency,
-            nomineeId,
-            nominatorId
-        );
+    ): Promise<MonetaryAmount<Currency<BTCUnit>, BTCUnit>> {
+        const totalGlobalReward = await this.computeRewardInRewardsPool(wrappedCurrency, vaultId);
```
nud3l, line 258: The `await`s here can also go into a `Promise.all`

nud3l (changes requested):
**Main Feedback**

Overall, this looks already quite good to me. The biggest point is the way that the wrapped currency is currently implemented. I commented that directly in the code, but after checking all the files, I find it more ergonomic to set the wrapped currency on a global level when instantiating the API rather to have it as an extra argument on several API calls.

As an example, this feels more natural to me:

```
const DEFAULT_PARACHAIN_ENDPOINT = "wss://kintsugi.interlay.io/api/parachain";
const wrappedAsset = KBTC;
const isMainnet = true;
const kBTC = await createInterbtcAPI(DEFAULT_PARACHAIN_ENDPOINT, wrappedAsset, isMainnet);
```

Especially since we will have first the kBTC and interBTC networks that anyway require a different network connection before we are working on other wrapped assets.

**Implementation Feedback**

There are a couple of places where you can execute network calls in parallel by using `map`/`Promise.all`. I am not sure I've marked all of them but good to have a look for those as they are the biggest impact on performance in the lib.

package.json
gregdhill: On naming, would it be nicer to properly camelCase some of these types? i.e. `DotAmount`

src/parachain/nomination.ts
gregdhill: I would also prefer this, there are far to many parameters anyway on some functions. For example, I found it difficult to workout what arguments were required for `issueSingle` when I was working on the tests. It would be nicer to adopt the builder pattern for these utils I think.

src/parachain/system.ts
```diff
@@ -52,4 +60,9 @@ export class DefaultSystemAPI implements SystemAPI {
         const head = await this.api.rpc.chain.getFinalizedHead();
         return await this.api.query.security.parachainStatus.at(head);
     }
+
+    async setCode(code: string): Promise<void> {
```
gregdhill, line 64: Maybe add a comment here that this delegates through `sudoUncheckedWeight`, otherwise it is misleading.

package.json
nud3l: I prefer the ticker-style naming. But that's personal preference

src/parachain/nomination.ts
savudani8: The builder pattern sounds good, in a different PR though

nud3l:
> TODO in a separate PR:
> 
>     * reproduce the removed `vaultRegistry.getTotalCollateralization` in the lib using storage values

Makes sense, we'll need this for the release of the lib.

>     * in the `vaultsAPI`, remove `collateralCurrency` parameter from methods and infer based on the set account

Since we changed the way multi-collateral is going to work, actually keeping the parameter is likely required.

src/parachain/issue.ts
```diff
@@ -206,29 +234,34 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI {
         }
     }
 
+    async craftRequestTx(
+        vaultId: AccountId,
+        amount: MonetaryAmount<WrappedCurrency, BitcoinUnit>
+    ): Promise<SubmittableExtrinsic<"promise">> {
+        const vault = await this.vaultsAPI.get(vaultId);
+        let griefingCollateral = await this.getGriefingCollateral(amount, vault.collateralCurrency);
```
nud3l, line 242: technically, it would help performance if the griefing collateral would be submitted as a parameter to this function as a mapping currencyId -> griefing collateral instead of fetching the griefing collateral rate on-chain for every vault.

src/parachain/oracle.ts
```diff
@@ -78,65 +82,71 @@ export interface OracleAPI extends TransactionAPI {
...
      * @returns The period of time (in milliseconds) after an oracle's last submission
      * during which it is considered online
      */
     getOnlineTimeout(): Promise<number>;
+    /**
+     * @param key A key defining an exchange rate or a BTC network fee estimate
+     * @returns Whether the oracle entr for the given key has been updated
+     */
+    getRawValuesUpdated(key: OracleKey): Promise<boolean>;
+    waitForFeeEstimateUpdate(): Promise<void>;
```
nud3l, line 97: Please add a description to those two new API functions.

nud3l:
@theodorebugnet @gregdhill please also have a look at that PR. I'd like to merge this soonish.

src/parachain/issue.ts
nud3l: You can do this in a later PR though!

src/types/vault.ts
```diff
@@ -14,22 +15,23 @@ export enum VaultStatusExt {
...
 }
 
-export interface SystemVaultExt {
-    toBeIssuedTokens: BTCAmount;
-    issuedTokens: BTCAmount;
-    toBeRedeemedTokens: BTCAmount;
+export interface SystemVaultExt<U extends BitcoinUnit> {
+    toBeIssuedTokens: MonetaryAmount<Currency<U>, U>;
+    issuedTokens: MonetaryAmount<Currency<U>, U>;
+    toBeRedeemedTokens: MonetaryAmount<Currency<U>, U>;
```
theodorebugnet, line 36:
```suggestion
export interface SystemVaultExt<WrappedUnit extends BitcoinUnit> {
    toBeIssuedTokens: MonetaryAmount<Currency<WrappedUnit>, WrappedUnit>;
    issuedTokens: MonetaryAmount<Currency<WrappedUnit>, WrappedUnit>;
    toBeRedeemedTokens: MonetaryAmount<Currency<WrappedUnit>, WrappedUnit>;
```

theodorebugnet (commented):
Looks very good

src/types/vault.ts
```diff
@@ -14,22 +15,23 @@ export enum VaultStatusExt {
     CommittedTheft,
 }
 
-export interface VaultExt {
+export interface VaultExt<U extends BitcoinUnit> {
```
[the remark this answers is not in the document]
savudani8: `WrappedUnit` might be misleading at the moment because they're all based on BTC, but I'll update since we're preparing to support new ones anyways

package.json
```diff
@@ -1,15 +1,17 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "0.27.1",
-  "description": "JavaScript library to interact with InterBTC",
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
savudani8: Good catch!

theodorebugnet (approved):

gregdhill (approved):

PR #331, opened 2021-09-30
feat: added function to fetch txid of the earliest payment to a recipient address, independent of utxo

theodorebugnet:


src/external/electrs.ts
```diff
@@ -224,11 +237,33 @@ export class DefaultElectrsAPI implements ElectrsAPI {
         return amount;
     }
 
+    async getEarliestPaymentToRecipientAddressTxId(recipientAddress: string, amount?: BitcoinAmount): Promise<string> {
+        try {
+            const txes = await this.getData(this.addressApi.getAddressTxHistory(recipientAddress));
+            if (txes.length >= 25) {
```
gregdhill, line 243: Hmm actually we may as well iterate over the past transactions to get the first.

src/external/electrs.ts
```diff
@@ -224,11 +237,33 @@ export class DefaultElectrsAPI implements ElectrsAPI {
...
+                    return oldestTx.txid;
+                }
+            }
+        } catch (e) {
+            return Promise.reject(new Error(`Error during tx lookup by address: ${e}`));
+        }
+        return Promise.reject(new Error("No transaction found for recipient and amount"));
+    }
+
     async getUtxoTxIdByRecipientAddress(recipientAddress: string, amount?: BitcoinAmount): Promise<string> {
```
gregdhill, line 262: Is this still used? I can't foresee where we would still need to fetch the UTXOs.

gregdhill (changes requested):

src/external/electrs.ts
theodorebugnet: Only issue is that the paging is kinda meh:
 - the first call returns the newest 25 confirmed txs, and then up to 50 mempool txs -> meaning have to filter for confirmed txs to know if we need paging
 - if there's exactly 25 txs, from the docs it sounds like this will return all 25 and then the next call will return []

Can definitely do it if it's necessary but given that we never expect to see it, not sure whether the added complexity is worth it.

gregdhill (approved):

PR #338, opened 2021-11-30
v1.3.0

savudani8:
- Uses metadata v14 along with minimal manual definitions
- Replicates RPC logic in TS. TBD if this logic is to be kept and applied on hydra data
- Comments out issue and redeem cancellation tests. These depend on waiting for request expiry which has to be implemented without depending on RPC
- Adds methods to the VaultExt object, similar to the approach on the chain
- Encodes `InterbtcPrimitivesVaultId` as a string in maps of form ``${id.accountId.toString()}-${wrappedIdLiteral}-${collateralIdLiteral}`` because creating the decoded type in the UI would be verbose. Decoding method is also added
- Removes unit tests that check the correctness of dependencies
- Uses docker images are the latest I could find that work together
- Upgrades to Yarn v2.0


src/parachain/issue.ts
```diff
@@ -207,17 +203,30 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI {
 
     async request(
         amount: MonetaryAmount<WrappedCurrency, BitcoinUnit>,
-        vaultId?: AccountId,
+        vaultAccountId?: AccountId,
```
nud3l, line 206: I think those two arguments should be combined into one argument that is an object. Because either you submit both or none, but not just one of those.

src/parachain/issue.ts
```diff
@@ -207,17 +203,30 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI {
...
+        collateralCurrencyIdLiteral?: CurrencyIdLiteral,
         atomic: boolean = true,
         retries: number = 0,
-        cachedVaults?: Map<AccountId, MonetaryAmount<WrappedCurrency, BitcoinUnit>>
+        cachedVaults?: Map<string, MonetaryAmount<WrappedCurrency, BitcoinUnit>>,
```
nud3l, line 210: I'm also wondering about the logic of this function. If a user has a vault selected the user would still need to send the cached vaults as well to prevent the function from loading those again.

Maybe the `cachedVaults` should be `selectedVaults` and if there is only a single element then this is the vault the user wants to issue with? Otherwise the `cachedVaults` and the `vaultAccountId` together with the `collateralCurrencyIdLiteral` are somewhat redundant arguments.

src/parachain/issue.ts
```diff
@@ -231,18 +240,28 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI {
...
     }
 
     async craftRequestTx(
-        vaultId: AccountId,
+        vaultId: InterbtcPrimitivesVaultId,
         amount: MonetaryAmount<WrappedCurrency, BitcoinUnit>
     ): Promise<SubmittableExtrinsic<"promise">> {
-        const vault = await this.vaultsAPI.get(vaultId);
+        const collateralId = currencyIdToLiteral(vaultId.currencies.collateral);
+        const vault = await this.vaultsAPI.get(vaultId.accountId, collateralId);
```
nud3l, line 264: Do we really need to query the entire vault object again? It seems like we only need this here:
```
let griefingCollateral = await this.getGriefingCollateral(amount, vault.collateralCurrency);
```

But the collateral currency is already assigned here `const collateralId = currencyIdToLiteral(vaultId.currencies.collateral);`.

src/parachain/issue.ts
```diff
@@ -255,14 +274,17 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI {
...
-        amountsPerVault: Map<AccountId, MonetaryAmount<WrappedCurrency, BitcoinUnit>>,
+        amountsPerVault: Map<string, MonetaryAmount<WrappedCurrency, BitcoinUnit>>,
         atomic: boolean
     ): Promise<Issue[]> {
         const txs = await Promise.all(
             Array.from(amountsPerVault.entries()).map(
                 ([vaultId, amount]) =>
                     new Promise<SubmittableExtrinsic<"promise">>((resolve) => {
-                        this.craftRequestTx(vaultId, amount).then((tx) => resolve(tx));
+                        this.craftRequestTx(
```
nud3l, line 284: `craftRequestTx` is doing two network reads in sequence and sends one transaction. In the worst case, this makes this request call very very slow. Can we somehow already submit the griefing collateral as part of making the tx so we get rid of the reads? Can be a future feature though as this is probably not super trivial/requires changes in the UI side of the code as well.

src/parachain/issue.ts
```diff
@@ -231,18 +240,28 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI {
...
-        vaultId: AccountId,
+        vaultId: InterbtcPrimitivesVaultId,
         amount: MonetaryAmount<WrappedCurrency, BitcoinUnit>
     ): Promise<SubmittableExtrinsic<"promise">> {
-        const vault = await this.vaultsAPI.get(vaultId);
+        const collateralId = currencyIdToLiteral(vaultId.currencies.collateral);
+        const vault = await this.vaultsAPI.get(vaultId.accountId, collateralId);
         let griefingCollateral = await this.getGriefingCollateral(amount, vault.collateralCurrency);
         // add() here is a hacky workaround for rounding errors
         const oneHundred = newMonetaryAmount(100, vault.collateralCurrency);
```
nud3l, line 267: Why is the griefing collateral in `Satoshi`? The griefing collateral is in DOT/KSM

src/parachain/redeem.ts
```diff
@@ -1,14 +1,18 @@
 import { ApiPromise } from "@polkadot/api";
 import { AddressOrPair, SubmittableExtrinsic } from "@polkadot/api/submittable/types";
-import { AccountId, Hash, H256, Header } from "@polkadot/types/interfaces";
+import { Hash, H256 } from "@polkadot/types/interfaces";
 import { EventRecord } from "@polkadot/types/interfaces";
-import { Bytes } from "@polkadot/types";
+import { Bytes, Option } from "@polkadot/types";
 import { Network } from "bitcoinjs-lib";
 import Big from "big.js";
```
nud3l, line 7: Do we need this dependency still?

src/parachain/redeem.ts
```diff
@@ -51,10 +62,10 @@ export interface RedeemAPI extends TransactionAPI {
     request(
         amount: MonetaryAmount<WrappedCurrency, BitcoinUnit>,
         btcAddressEnc: string,
-        vaultId?: AccountId,
+        vaultId?: InterbtcPrimitivesVaultId,
         atomic?: boolean,
         retries?: number,
-        availableVaults?: Map<AccountId, MonetaryAmount<WrappedCurrency, BitcoinUnit>>
+        availableVaults?: Map<string, MonetaryAmount<WrappedCurrency, BitcoinUnit>>
```
nud3l, line 68: Similar to issue, not sure we would want to have only one option: submit an array or the vaultId.

src/parachain/replace.ts
```diff
@@ -109,56 +120,51 @@ export class DefaultReplaceAPI extends DefaultTransactionAPI implements ReplaceA
...
         const vaultAccount = this.getAccount();
         if (vaultAccount === undefined) {
             return Promise.reject("Vault account must be set in the replace API");
         }
-        const vaultId = isKeyringPair(vaultAccount) ? vaultAccount.address : vaultAccount.toString();
-
-        const vault = await this.vaultsAPI.get(newAccountId(this.api, vaultId));
+        const vaultAccountId = isKeyringPair(vaultAccount) ? vaultAccount.address : vaultAccount.toString();
+        const vaultId = newVaultId(this.api, vaultAccountId, collateralCurrency, this.wrappedCurrency);
+        const vault = await this.vaultsAPI.get(
```
nud3l, line 144: Since the `collateralCurrency` is submitted as an arg, I think we don't need to load the entire vault object again.

src/parachain/replace.ts
```diff
@@ -244,21 +265,25 @@ export class DefaultReplaceAPI extends DefaultTransactionAPI implements ReplaceA
...
+    async mapReplaceRequests(vaultAccountId: AccountId): Promise<ReplaceRequestExt[]> {
         try {
-            const oldVaultReplaceRequests: [H256, ReplaceRequest][] =
-                await this.api.rpc.replace.getOldVaultReplaceRequests(vaultId);
-            const oldVaultReplaceRequestsExt = await this.parseRequestsAsync(oldVaultReplaceRequests);
-            const newVaultReplaceRequests: [H256, ReplaceRequest][] =
-                await this.api.rpc.replace.getNewVaultReplaceRequests(vaultId);
-            const newVaultReplaceRequestsExt = await this.parseRequestsAsync(newVaultReplaceRequests);
-            return new Map([...oldVaultReplaceRequestsExt, ...newVaultReplaceRequestsExt]);
+            const oldVaultReplaceRequests = await this.getOldVaultReplaceRequests(vaultAccountId);
```
nud3l, line 270: Those can be wrapped in a `Promise.all`

src/parachain/rewards.ts
```diff
@@ -9,244 +10,384 @@ import {
...
     /**
      * @param vaultId The account ID of the staking pool nominee
      * @param nominatorId The account ID of the staking pool nominator
+     * @param collateralCurrencyId Collateral currency used by the vault
      * @returns A Monetary.js amount object, representing the reward in the given currency
      */
     computeRewardInStakingPool(
-        vaultId: string,
-        nominatorId: string
+        vaultAccountId: AccountId,
```
nud3l, line 34: General comment on the function interfaces here: the ordering of the arguments should be consistent. Here, the accountId, collateralCurrency, nominator et.c are switched around between different interfaces. I suggest you stick with something like:
1. VaultId
2. collateralCurrency
3. NominatorId
4. ...

src/parachain/rewards.ts
```diff
@@ -9,244 +10,384 @@ import {
...
+        collateralCurrencyId: CollateralIdLiteral,
+        nonce?: number
     ): Promise<MonetaryAmount<Currency<BitcoinUnit>, BitcoinUnit>> {
-        const currencyId = tickerToCurrencyIdLiteral(this.wrappedCurrency.ticker);
+        const wrappedCurrencyId = tickerToCurrencyIdLiteral(this.wrappedCurrency.ticker) as WrappedIdLiteral;
         const [stake, rewardPerToken, rewardTally] = await Promise.all([
-            this.getStakingPoolStake(currencyId, vaultId, nominatorId),
-            this.getStakingPoolRewardPerToken(currencyId, vaultId),
-            this.getStakingPoolRewardTally(currencyId, vaultId, nominatorId),
+            this.getStakingPoolStake(collateralCurrencyId, vaultAccountId, nominatorId, nonce),
```
nud3l, line 167: This is what I mean with consistent ordering, it's a bit confusing when the order changes between different calls.

src/parachain/tokens.ts
```diff
@@ -103,8 +107,12 @@ export class DefaultTokensAPI extends DefaultTransactionAPI implements TokensAPI
 
     async transfer<U extends CurrencyUnit>(destination: string, amount: MonetaryAmount<Currency<U>, U>): Promise<void> {
         const amountSmallDenomination = this.api.createType("Balance", amount.toString());
```
nud3l, line 109: Naming convention here. I think in some other places you named this `amountAtomicUnit`. Good to stick with one name for the same thing.

src/parachain/transaction.ts
```diff
@@ -40,14 +40,26 @@ export class DefaultTransactionAPI {
...
                 .catch((error) => reject(error));
 
-            function callback(callbackObject: { unsubscribe: () => void; result: ISubmittableResult }): void {
+            function callback(api: ApiPromise, callbackObject: { unsubscribe: () => void; result: ISubmittableResult }): void {
                 const status = callbackObject.result.status;
                 if (status.isFinalized) {
+                    const dispatchError = callbackObject.result.dispatchError;
+                    if (dispatchError) {
+                        if (dispatchError.isModule) {
+                            // for module errors, we have the section indexed, lookup
```
nud3l, line 54: Nice! This will help a lot with debugging :)

src/parachain/vaults.ts
```diff
@@ -300,24 +328,39 @@ export class DefaultVaultsAPI extends DefaultTransactionAPI implements VaultsAPI
...
+        vaultAccountId: AccountId,
+        collateralCurrencyIdLiteral: CollateralIdLiteral
+    ): Promise<VaultExt<BitcoinUnit>> {
+        try {
+            const head = await this.api.rpc.chain.getFinalizedHead();
+            const collateralCurrency = currencyIdToMonetaryCurrency(
+                newCurrencyId(this.api, collateralCurrencyIdLiteral)
+            ) as CollateralCurrency;
+            const vaultId = newVaultId(this.api, vaultAccountId.toString(), collateralCurrency, this.wrappedCurrency);
+            const api = await this.api.at(head);
```
nud3l, line 345: Not sure I fully understand this, don't you have an `api` instance already?

src/types/vault.ts
```diff
@@ -28,6 +34,114 @@ export interface VaultExt<WrappedUnit extends BitcoinUnit> {
...
+            backedTokens as unknown as MonetaryAmount<Currency<BitcoinUnit>, BitcoinUnit>,
+            currencyIdToMonetaryCurrency(this.id.currencies.collateral) as Currency<CollateralUnit>
+        );
+        const secureCollateralThreshold = await this.getSecureCollateralThreshold();
+        const usedCollateral = backedTokensInCollateral.mul(secureCollateralThreshold);
+        const totalCollateral = await this.computeBackingCollateral();
+        return totalCollateral.min(usedCollateral);
+    }
+
+    // TODO: The functions below are duplicated from other APIs. Remove them after APIs
```
nud3l, line 117: Is this something that you want to do in this PR or later?

test/integration/parachain/release/redeem.test.ts
```diff
@@ -94,35 +113,35 @@ describe("redeem", () => {
...
-                });
-            });
-            await redeemAPI.cancel(redeemRequest.id.toString(), true);
-            const redeemRequestAfterCancellation = await redeemAPI.getRequestById(redeemRequest.id);
-            assert.isTrue(redeemRequestAfterCancellation.status === RedeemStatus.Reimbursed, "Failed to cancel issue request");
-            // Set issue period back to its initial value to minimize side effects.
-            await sudo(redeemAPI, (api) => api.setRedeemPeriod(initialRedeemPeriod));
-        });
-    }).timeout(5 * 60 * 1000);
+    // TODO: Uncomment after `subscribeToRedeemExpiry` is reimplemented
```
nud3l, line 116: For this or another PR?

test/integration/parachain/staging/issue.test.ts
```diff
@@ -188,57 +189,50 @@ describe("issue", () => {
...
         const requestLimits = await issueAPI.getRequestLimits();
-        assert.isTrue(requestLimits.singleVaultMaxIssuable.gt(InterBtcAmount.from.BTC(100)), "singleVaultMaxIssuable is not greater than 100");
+        assert.isTrue(requestLimits.singleVaultMaxIssuable.gt(InterBtcAmount.from.BTC(0.001)), "singleVaultMaxIssuable is not greater than 100");
         assert.isTrue(
             requestLimits.totalMaxIssuable.gt(requestLimits.singleVaultMaxIssuable),
             "totalMaxIssuable is not greater than singleVaultMaxIssuable"
         );
     });
 
+    // TODO: Uncomment after `subscribeToIssueExpiry` is reimplemented
```
nud3l, line 199: For this or another PR?

test/unit/utils/bitcoin.test.ts
```diff
@@ -23,12 +23,4 @@ describe("Bitcoin", () => {
             p2wpkhv0: "0xe8df018c7e326cc253faac7e46cdc51e68542c42",
         });
     });
-
-    it("should correctly decode & encode address", () => {
```
nud3l, line 27: Why don't we need this anymore?

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: "node-modules"
```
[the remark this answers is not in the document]
savudani8: Yes, it's for the upgrade of Yarn. It's present in https://github.com/polkadot-js/api too for instance

src/parachain/issue.ts
```diff
@@ -41,12 +52,13 @@ export interface IssueAPI extends TransactionAPI {
      * parachain (incurring an extra request).
      * @returns An object of type {singleVault, maxTotal, vaultsCache}
      */
-    getRequestLimits(vaults?: Map<AccountId, MonetaryAmount<WrappedCurrency, BitcoinUnit>>): Promise<IssueLimits>;
+    getRequestLimits(vaults?: Map<string, MonetaryAmount<WrappedCurrency, BitcoinUnit>>): Promise<IssueLimits>;
```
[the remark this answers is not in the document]
savudani8: I was thinking that `InterbtcPrimitivesVaultId` would be tricky to construct from the UI, but I think it will actually be fine. I'll use the strong type

src/parachain/issue.ts
savudani8: Given this reasoning, we might as well require the full `InterbtcPrimitivesVaultId` ( the tuple `[vaultAccountId, collateral, wrapped]`) instead of setting the wrapped currency in the API. If the UI is to know which vault has issuable tokens, it probably knows the full ID too.

src/parachain/issue.ts
savudani8: If a user specifies a `vaultAccountId`, the function returns without fetching any vaults. If none is specified, the `cachedVaults` array is used to select vaults for the issue request. In the absence of `cachedVaults`, the vault list is queried from the parachain. I think there's no redundance? As above, `vaultAccountId` and `collateralCurrencyIdLiteral` could be merged into `InterbtcPrimitivesVaultId`, which also includes the wrapped currency.

src/parachain/refund.ts
```diff
@@ -72,44 +61,30 @@ export class DefaultRefundAPI extends DefaultTransactionAPI implements RefundAPI
...
-            return parseRefundRequest(keyValuePair[1], this.btcNetwork, this.wrappedCurrency);
-        } catch (error) {
-            return Promise.reject(new Error(`Error fetching refund request by issue id: ${error}`));
+    async getRequestByIssueId(issueId: string): Promise<RefundRequestExt> {
+        const list = await this.list();
+        const requests = list.filter((request) => request.issueId === issueId);
+        if (requests.length !== 1) {
+            return Promise.reject(new Error(`No refund request fund for issue ID ${issueId}`));
         }
+        return requests[0];
```
tomjeatt, line 88: I'm a little confused by this - we're filtering the list, which could return multiple entries, then throwing an error if the result is longer than 1. What are the circumstances where we could get more than one result? e.g. could there be multiple matches for an issueId? If this is possible, then I would expect to throw two errors: one for no match, and one for multiple matches. If there can only ever be a single match, then I'd prefer to use `list.find`, as it makes the code a bit clearer (i.e. by using `find`, it's implicit that we're looking for a single item; by using `filter` the implication is that we're looking for 1 or more items).

I am looking at this function in isolation though, so there could well be some context I'm missing :)

src/parachain/tokens.ts
```diff
@@ -66,17 +66,17 @@ export class DefaultTokensAPI extends DefaultTransactionAPI implements TokensAPI
...
     async balanceLocked<U extends CurrencyUnit>(
         currency: Currency<U>,
         id: AccountId
     ): Promise<MonetaryAmount<Currency<U>, U>> {
-        const currencyId = tickerToCurrencyIdLiteral(currency.ticker);
-        const account = await this.api.query.tokens.accounts(id, currencyId);
+        const currencyIdLiteral = tickerToCurrencyIdLiteral(currency.ticker);
+        const account = await this.api.query.tokens.accounts(id, newCurrencyId(this.api, currencyIdLiteral));
         return newMonetaryAmount(account.reserved.toString(), currency);
     }
```
tomjeatt, line 81: Minor, but could this duplicated code be consolidated? Other than the returned value these two methods are the same.

src/types/currency.ts
```diff
@@ -85,3 +94,12 @@ export function currencyIdToMonetaryCurrency<U extends CurrencyUnit>(currencyId:
     }
     throw new Error("No CurrencyId entry for provided ticker");
 }
+
+export function currencyIdToLiteral(currencyId: InterbtcPrimitivesCurrencyId): CurrencyIdLiteral {
+    return tickerToCurrencyIdLiteral(currencyIdToMonetaryCurrency(currencyId).ticker);
```
tomjeatt, line 99: Minor, but I find this quite hard to read! Would be a bit easier to follow if `currencyIdToMonetaryCurrency(currencyId).ticker` was assigned to a variable rather than invoked directly in the function.

src/types/vault.ts
```diff
@@ -28,6 +34,114 @@ export interface VaultExt<WrappedUnit extends BitcoinUnit> {
...
+    async getIssuableTokens(): Promise<MonetaryAmount<Currency<WrappedUnit>, WrappedUnit>> {
+        const isBanned = await this.isBanned();
+        if (isBanned) {
+            return newMonetaryAmount(0, currencyIdToMonetaryCurrency(this.id.currencies.wrapped));
+        }
+        const freeCollateral = await this.getFreeCollateral();
+        const secureCollateralThreshold = await this.getSecureCollateralThreshold();
+        const oracleAPI = new DefaultOracleAPI(this.api, currencyIdToMonetaryCurrency(this.id.currencies.wrapped));
+        const backableWrappedTokens = await oracleAPI.convertCollateralToWrapped(freeCollateral);
+        // Force type-assert here as the oracle API only uses wrapped Bitcoin
```
tomjeatt, line 80: Nice comment :)

src/utils/issueRedeem.ts
```diff
@@ -185,21 +201,22 @@ export async function redeem(
...
     redeemAPI.setAccount(redeemingAccount);
 
     const btcAddress = "bcrt1qujs29q4gkyn2uj6y570xl460p4y43ruayxu8ry";
     const [redeemRequest] = await redeemAPI.request(
         amount,
         btcAddress,
-        vaultAddress ? newAccountId(api, vaultAddress) : undefined,
+        vaultId,
         atomic,
         0 // retries
```
tomjeatt, line 221: Minor, but if you need to comment this to describe what it is then it might be cleaner to declare it as a variable, i.e. `const retries = 0`.

test/integration/parachain/release/redeem.test.ts
```diff
@@ -94,35 +113,35 @@ describe("redeem", () => {
...
+    // it("should cancel a redeem request", async () => {
```
tomjeatt, line 117: Could you skip this instead of commenting out? That way it should show up in the test runner as a skipped test which would make it visible (and more likely to be remembered!).

test/integration/parachain/staging/issue.test.ts
```diff
@@ -188,57 +189,50 @@ describe("issue", () => {
...
-
-                // Wait for issue expiry callback
-                await new Promise<void>((resolve, _) => {
-                    issueAPI.subscribeToIssueExpiry(newAccountId(api, userAccount.address), (requestId) => {
-                        if (stripHexPrefix(requestResult.id.toString()) === stripHexPrefix(requestId.toString())) {
-                            resolve();
-                        }
-                    });
-                });
+    // it("should cancel an issue request", async () => {
```
tomjeatt, line 201: As above, could this be skipped rather than commented out?

test/integration/parachain/staging/issue.test.ts
```diff
@@ -188,57 +189,50 @@ describe("issue", () => {
...
-
-    it("should list issue request by a vault", async () => {
-        const vaultToBanAddress = vault_to_ban.address;
-        const vaultToBanId = api.createType("AccountId", vaultToBanAddress);
-        const issueRequests = await issueAPI.mapIssueRequests(vaultToBanId);
-        issueRequests.forEach((request) => {
-            assert.deepEqual(request.vaultParachainAddress, vaultToBanAddress);
-        });
-    });
+    //             const issueRequest = await issueAPI.getRequestById(requestResult.id);
```
tomjeatt, line 224: As above, could this be skipped rather than commented out?

test/unit/interbtc-api.test.ts
```diff
@@ -13,7 +15,7 @@ describe("InterBTCAPI", () => {
     before(async () => {
         const api = sinon.createStubInstance(ApiPromise);
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
tomjeatt, line 17: If we need to disable linting, I quite like having a comment explaining why :)

tomjeatt (approved):
Looks good :) few comments, all minor. No blockers so happy to approve.

src/parachain/issue.ts
savudani8: If you mean two calls to compute the griefing collateral that's a good point, I changed it to call in parallel

src/parachain/redeem.ts
savudani8: Yes, for rates such as `getFeeRate()`

src/parachain/rewards.ts
savudani8: If we merge the vaultId and collateralCurrency into an `InterbtcPrimitivesVaultId`, the interface would be simpler

src/parachain/vaults.ts
savudani8: Queries such as `api.query.vaultRegistry.vaults.at(...)` have become deprecated. We first need to get an `api` object for a specific block and then query. I tried it here to see if it works, but I can keep it uniform with the rest until I change everywhere

src/types/vault.ts
savudani8: It's for later

test/integration/parachain/release/redeem.test.ts
savudani8: Later PR

test/integration/parachain/staging/issue.test.ts
savudani8: Later PR

test/unit/utils/bitcoin.test.ts
savudani8: It cannot be run as a unit test because when `btcAddressFromParams` attempts to construct a `BitcoinAddress` on an empty `registry`, it will fail because no metadata is set. This functionality is already tested in the replace integration tests though.

savudani8: It cannot be run as a unit test because when btcAddressFromParams attempts to construct a BitcoinAddress on an empty registry, it will fail because no metadata is set. This functionality is already tested in the replace integration tests though.

src/parachain/refund.ts
savudani8: Good advice, thanks!

src/parachain/tokens.ts
savudani8: One method returns the `reserved` while the other returns `free` balance. Good point though, I've refactored.

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.8"
 services:
     interbtc:
-        image: "interlayhq/interbtc:interbtc-standalone-1.0.3"
+        image: "interlayhq/interbtc:interbtc-standalone-master-9ec56a59-1636982230"
```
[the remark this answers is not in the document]
gregdhill: @savudani8 I think these are tagged now right? Can we update this?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.3.0",
```
gregdhill, line 3: Was `1.3.0` previously published?

src/factory.ts
```diff
@@ -23,20 +23,19 @@ export function createProvider(endpoint: string, autoConnect?: number | false |
...
     const provider = createProvider(endpoint, autoConnect);
-    const types = getAPITypes();
-    const rpc = getRPCTypes();
-    return ApiPromise.create({ provider, types, rpc });
+    return ApiPromise.create({ provider });
 }
 
 export async function createInterbtcAPI(
     endpoint: string,
+    nativeCurrency: CollateralCurrency,
```
gregdhill, line 31:
```suggestion
    collateralCurrency: CollateralCurrency,
```

gregdhill: native == KINT
collateral == KSM
wrapped = KBTC

src/interbtc-api.ts
```diff
@@ -77,25 +77,40 @@ export class DefaultInterBTCAPI implements InterBTCAPI {
 
     constructor(
         readonly api: ApiPromise,
+        nativeCurrency: CollateralCurrency,
```
gregdhill, line 80:
```suggestion
        collateralCurrency: CollateralCurrency,
```

src/parachain/fee.ts
```diff
@@ -55,10 +61,24 @@ export class DefaultFeeAPI implements FeeAPI {
 
     async getGriefingCollateral<C extends CollateralUnit>(
         amount: MonetaryAmount<Currency<BitcoinUnit>, BitcoinUnit>,
-        griefingCollateralRate: Big,
-        collateralCurrency: Currency<C>
+        collateralCurrency: Currency<C>,
+        type: GriefingCollateralType
```
gregdhill, line 65: What is the benefit of have one function? IMO this would be cleaner with `getIssueGriefing` and `getReplaceGriefing`.

src/parachain/issue.ts
gregdhill: Let's do whatever is faster to implement, bear in mind that for the time being setting the wrapped currency ID will have no effect - we only support one per chain.

src/parachain/issue.ts
```diff
@@ -207,16 +192,29 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI {
...
         retries: number = 0,
-        cachedVaults?: Map<AccountId, MonetaryAmount<WrappedCurrency, BitcoinUnit>>
+        cachedVaults?: Map<InterbtcPrimitivesVaultId, MonetaryAmount<WrappedCurrency, BitcoinUnit>>,
     ): Promise<Issue[]> {
         try {
-            if (vaultId) {
+            if (vaultAccountId) {
+                if (!collateralCurrencyIdLiteral) {
+                    return Promise.reject(
+                        new Error("A collateral currency must be specified along with the vault account ID")
```
gregdhill, line 205: Not sure what is best-practice (maybe @tomjeatt has a recommendation) but I'm not a huge fan of in-line error messages.

src/parachain/nomination.ts
```diff
@@ -287,37 +326,44 @@ export class DefaultNominationAPI extends DefaultTransactionAPI implements Nomin
...
-    async getNominationStatus(nominatorId: string, vaultId: string): Promise<NominationStatus> {
+    async getNominationStatus(nominatorId: AccountId, vaultId: AccountId): Promise<NominationStatus> {
         // There is at most one entry determined by this pair
         const filteredNominations = await this.getFilteredNominations(nominatorId, vaultId);
         if (filteredNominations.length === 0) {
             return Promise.reject(new Error("No nomination associated with this (nominator, vault) pair"));
         }
-        const filteredNomination = filteredNominations[0];
-        if (filteredNomination[1].isZero()) {
+        const [, , , nominatedAmount] = filteredNominations[0];
```
gregdhill, line 365: Gotta say this is pretty ugly and likely very error-prone...

src/parachain/nomination.ts
```diff
@@ -336,21 +382,29 @@ export class DefaultNominationAPI extends DefaultTransactionAPI implements Nomin
         }
         const zero = newMonetaryAmount<CollateralUnit>(0, collateralCurrency as Currency<CollateralUnit>);
         return filteredNominations
-            .filter(
-                (v) => v[2].ticker === this.wrappedCurrency.ticker && v[1].currency.ticker === collateralCurrency.ticker
-            )
-            .map((v) => v[1])
+            .filter((v) => {
+                const [, vaultId, , ,] = v;
```
gregdhill, line 386: Same here, there must be a way to have stronger types.

src/parachain/transaction.ts
gregdhill: Yeah I'm also really happy with this!

gregdhill (changes requested):

src/parachain/issue.ts
tomjeatt: Good point - libs aren't something I've had much experience with, but in a client app I would use a dictionary for handling errors.

package.json
savudani8: Seems so https://github.com/interlay/interbtc-api/releases/tag/1.3.0

src/parachain/fee.ts
savudani8: Those functions do exist, but they're each exposed in their respective API. I added this function to parallelize two calls that were previously made in sequence

src/parachain/issue.ts
savudani8: We have quite a few of these (here too https://github.com/interlay/interbtc-api/blob/master/src/clients/client.ts#L51). I'll add to backlog

gregdhill (approved):

PR #346, opened 2022-01-18
feat(escrow): Extrinsics and getters

savudani8:
- Adds the escrow API that allows for locking and withdrawing governance currencies. `withdraw` was not tested because it would require waiting for one week, and instant-seal is not yet available to speed up the test.
- Adds missing `withdrawRewards` extrinsic for staking rewards
- Reimplements the on-chain logic for `balance_at` and `total_supply`, to avoid making an RPC call to the chain. These methods have been kept non-async so that hydra or `index` can compute results without incurring network delays


src/parachain/escrow.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    async totalVotingSupply(
+        blockNumber?: number
+    ): Promise<MonetaryAmount<Currency<VoteUnit>, VoteUnit>> {
+        const [currentBlockNumber, epoch, span, rawSlopeChanges] = await Promise.all([
+            this.systemAPI.getCurrentBlockNumber(),
+            this.api.query.escrow.epoch(),
+            this.getSpan(),
+            this.api.query.escrow.slopeChanges.entries()
```
gregdhill, line 150: This may grow very big IIRC.

src/parachain/escrow.ts
```diff
@@ -0,0 +1,243 @@
...
+                break;
+            }
+
+            last_point.slope.saturating_accrue(d_slope);
+            last_point.ts = t_i;
+        }
+
+        last_point.bias
+    }
+    */
```
gregdhill, line 195: Remove and add permalink.

src/parachain/escrow.ts
savudani8: In practice this wouldn't be called at all. What I did was to separate the data fetching from the logic so `index` can reuse the logic. Then, the UI would call index exclusively. I'll add a doc `@remark` about the performance of this getter though.

gregdhill (approved):

PR #348, opened 2022-01-19
feat: griefing collateral uses native currency

savudani8:
Griefing collateral is computed in the native currency, by querying the `vaultRegistry.getGriefingCollateralCurrencyId` constant


docker-compose.yml
```diff
@@ -66,31 +66,16 @@ services:
...
-    faucet:
-        image: "interlayhq/interbtc-clients:faucet-1-5-0"
+        image: "interlayhq/interbtc-clients:oracle-standalone-metadata-1-5-4"
         command:
             - /bin/sh
             - -c
             - |
-                echo '{"bob_stash": "0x1a7d114100653850c65edecda8a9b2b4dd65d900edef8e70b1a6ecdcda967056"}' > keyfile.json
-                faucet --keyfile="keyfile.json" --keyname=bob_stash --btc-parachain-url 'ws://interbtc:9944' --user-allowance 1 --vault-allowance 500 --http-addr '[::0]:3036'
+                oracle-standalone-metadata --keyring charlie --btc-parachain-url 'ws://interbtc:9944' --currency-id DOT KSM KINT
```
nud3l, line 74: I'm a bit surprised that the integration test is passing. The vaults seem to be set to use `DOT` collateral and issue `interBTC`. In this case, the currency of the griefing collateral should be `INTR` which is not submitted in the oracle here.

I think the change that you made @savudani8 generalizes, so the lib code should work. My surprise is just that I would have expected a different behavior on the parachain. Can you confirm this @sander2 and @savudani8?

savudani8: Checked now, it passes because the native currency is set to DOT on standalone

sander2 (approved):

PR #349, opened 2022-01-20
Refactoring

savudani8:
Changes:
- unidirectional API dependencies
- test cleanup - no more individual API initializations, but using the whole `interBtcAPI` everywhere. This helps new users easily understand how to use this lib by peeking at the tests
- the test cleanup emphasized that some parameters are passed unnecessarily, which are removed

For a future PR:
- clears the way for using a single `transactionAPI` object per API. This helps with the upcoming feature of notifying the UI of wallet events without adding duplicated code in each API. Rather than extend a `TransactionAPI` class, each API will be initialized the same `transactinAPI` object, passed by reference. Another advantage of this approach is being able to change the account in all APIs at once without re-initializing anything. To avoid the disadvantage of a counter-intuitive interface caused by passing by reference, this object will be private and only changeable through setters.


src/factory.ts
```diff
@@ -28,14 +28,13 @@ export function createPolkadotAPI(endpoint: string, autoConnect?: number | false
...
     endpoint: string,
-    collateralCurrency: CollateralCurrency,
     network: BitcoinNetwork = "mainnet",
     wrappedCurrency: WrappedCurrency = InterBtc,
     account?: AddressOrPair,
     autoConnect?: number | false | undefined
 ): Promise<InterBTCAPI> {
     const api = await createPolkadotAPI(endpoint, autoConnect);
-    return new DefaultInterBTCAPI(api, collateralCurrency, network, wrappedCurrency, account);
+    return new DefaultInterBTCAPI(api, network, wrappedCurrency, account);
```
nud3l, line 37: In the future, we should also inject the wrappedCurrency at runtime, but not in this PR.

src/parachain/issue.ts
```diff
@@ -286,9 +290,14 @@ export class DefaultIssueAPI extends DefaultTransactionAPI implements IssueAPI {
         }
     }
 
-    async execute(requestId: string, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void> {
+    async execute(requestId: string, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes, electrsAPI?: ElectrsAPI): Promise<void> {
```
nud3l, line 293: If I understand this correctly, it would not be possible to do `issue.execute("requestId")` directly even though the other parameters are optional. IMO that is not a good change since it's confusing that optional parameters are required for the function to complete successfully and it requires an extra step for whoever is consuming this API.

What's the reason for removing electrs here? I'd prefer to keep this in this function so that the execute can be done as simple as possible.

src/parachain/redeem.ts
```diff
@@ -256,15 +252,16 @@ export class DefaultRedeemAPI extends DefaultTransactionAPI implements RedeemAPI
         }
     }
 
-    async execute(requestId: string, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void> {
+    async execute(requestId: string, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes, electrsAPI?: ElectrsAPI): Promise<void> {
```
nud3l, line 255: Similar issue as with `issue.execute()`

src/parachain/replace.ts
```diff
@@ -189,9 +182,9 @@ export class DefaultReplaceAPI extends DefaultTransactionAPI implements ReplaceA
         await this.sendLogged(requestTx, this.api.events.replace.AcceptReplace);
     }
 
-    async execute(requestId: string, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void> {
-        const parsedRequestId = this.api.createType("H256", "0x" + requestId);
-        [merkleProof, rawTx] = await getTxProof(this.electrsAPI, btcTxId, merkleProof, rawTx);
+    async execute(requestId: string, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes, electrsAPI?: ElectrsAPI,): Promise<void> {
```
nud3l, line 185: Similar issue as `issue.execute()`. Possibly a workaround could be to have the correct electrs URL defined as part of the bitcoin network and define `getTxProof` as follows:

`getTxProof(btcTxId, electrsAPI?)`
-> if only the btxTxId is given, it'll fallback to the default URL of the connected BTC testnet. If electrsAPI is given it uses that.
-> The `merkleProof` and `rawTx` parameter seem necessary, why do we need those?

src/parachain/vaults.ts
```diff
@@ -245,38 +244,63 @@ export interface VaultsAPI extends TransactionAPI {
...
+     * @param electrsAPI API used to fetch the merkleProof and rawTx,
+     * if these paramters are not provided
      * @param btcTxId The hash of the transaction
      * @param merkleProof The proof of tx inclusion.
      * @param rawTx The raw Bitcoin transaction.
      */
-    reportVaultTheft(vaultId: AccountId, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void>;
+    reportVaultTheft(
+        vaultAccountId: AccountId,
+        electrsAPI?: ElectrsAPI,
```
nud3l, line 255: Similar issues as with `issue.execute()`

src/utils/bitcoin.ts
```diff
@@ -83,7 +83,7 @@ export function decodeBtcAddress(
 }
 
 export async function getTxProof(
```
nud3l, line 85: I think this function can be simplified:
1. Require `btcTxId` as a parameter then you don't need the extra if condition and throw an error. It's just required on function call. If you already have both the merkleProof and rawTx, you'd anyway not call that function.
2. Default to some electrsAPI OR make it a required parameter. Having an optional parameter that just throws if not supplied is bad UX for the API.

nud3l (commented):
I think I overall might have missed something related to that electrsAPI change. As far as I understand it, the current change would require manual submission of the electrsAPI on issue, redeem, replace execution and might fail if the optional parameter is not provided. I've left detailed comments below so let me know if I misunderstood that change.

.env.staging
```diff
@@ -24,5 +24,5 @@ export BITCOIN_CORE_PORT="18443";
 export BITCOIN_CORE_WALLET="Alice";
 export SUDO_URI="//Alice";
 
-const NATIVE_CURRENCY_TICKER="DOT";
+const COLLATERAL_CURRENCY_TICKER="DOT";
```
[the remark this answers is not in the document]
savudani8: Added a ticket, I agree it would simplify deployment of the entire JS stack

src/utils/bitcoin.ts
savudani8: I refactored this to take the `TxFetchingDetails` type, let me know what you think

src/types/bitcoinTypes.ts
```diff
@@ -1 +1,22 @@
...
+
 export type BitcoinNetwork = "mainnet" | "testnet" | "regtest";
+export type TxInclusionDetails = {
+    merkleProof: Bytes;
+    rawTx: Bytes;
+}
+export type TxInclusionDeterminants = {
+    btcTxId: string,
+}
+export type TxFetchingDetails = TxInclusionDetails | TxInclusionDeterminants;
```
gregdhill, line 11: Not sure about the naming of some of these types, but can't really think of an alternative. At the very least `TxInclusionDeterminants` probably doesn't need to be a complex type.

src/parachain/issue.ts
```diff
@@ -89,11 +90,9 @@ export interface IssueAPI extends TransactionAPI {
      * @remarks If `txId` is not set, the `merkleProof` and `rawTx` must both be set.
      *
      * @param issueId The ID returned by the issue request transaction
-     * @param txId (Optional) The ID of the Bitcoin transaction that sends funds to the vault address.
-     * @param merkleProof (Optional) The merkle inclusion proof of the Bitcoin transaction.
-     * @param rawTx (Optional) The raw bytes of the Bitcoin transaction
+     * @param txFetcher TxFetchingDetails object
      */
-    execute(issueId: string, txId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void>;
+    execute(requestId: string, txFetchingDetails: TxFetchingDetails): Promise<void>;
```
nud3l, line 95: I'm wondering if we shouldn't just simplify the API at this point by requiring the btcTxId, fetch the proof ourselves and if some really wants to dig deep, we could in a later sprint introduce `executeAdvanced` with extra stuff to set. But at this point we are polluting the API interface for a use case that does not exist (maybe yet) in practice.
Another consideration for me is that this change will have breaking changes on the UI without good reason really.


So my vote is to simplify all of these to only take the `btcTxId` as a required param and that's it. I know that is a change from what we tried to do before, but I think better build things where there is demand for and make our own life easier at this point by reducing complexity.

What do you think? Also @gregdhill do you have an opinion on this?

src/parachain/vaults.ts
```diff
@@ -245,38 +244,57 @@ export interface VaultsAPI extends TransactionAPI {
      * @remarks If `txId` is not set, the `merkleProof` and `rawTx` must both be set.
      *
      * @param vaultId The account of the vault to check.
-     * @param btcTxId The hash of the transaction
-     * @param merkleProof The proof of tx inclusion.
-     * @param rawTx The raw Bitcoin transaction.
+     * @param txFetcher TxFetchingDetails object
      */
-    reportVaultTheft(vaultId: AccountId, btcTxId?: string, merkleProof?: Bytes, rawTx?: Bytes): Promise<void>;
+     reportVaultTheft(
```
nud3l, line 249: Formatting

src/types/bitcoinTypes.ts
```diff
@@ -1 +1,19 @@
+import { Bytes } from "@polkadot/types";
```
nud3l, line 1: If we just require the btcTxId on `execute`, we don't need those involved types.

src/parachain/issue.ts
gregdhill: I agree, definitely doesn't make sense to increase complexity for the sake of non-existent downstream users. We've already had this bite us with the `interbtc-api` / `interbtc-js` split.

PR #350, opened 2022-01-20
Chore: error handling in getPremiumRedeemVaults

ns212:


savudani8 (approved):
You'll need to increase the version in package.json as well

sander2 (changes requested):
It's good not to fatally crash when an error occurs, but I feel like this isn't really addressing the root of the issue - why is `isBelowPremiumThreshold` producing an error in the first place?

savudani8:
@sander2 I think this can be merged since the root problem was in `index`

sander2 (approved):
Now that the root issue has been identified and fixed, I no longer have any objections here

PR #352, opened 2022-01-25
feat: add findBlockEpoch implementation

gregdhill:

### DIFF
--- a/src/parachain/escrow.ts
+++ b/src/parachain/escrow.ts
@@ -225,37 +225,88 @@ export class DefaultEscrowAPI implements EscrowAPI {
         return newMonetaryAmount(rawBalance.toString(), toVoting(this.governanceCurrency));
     }
 
-    private rawBalanceAt(escrowPoint: RWEscrowPoint, height: number): BN {
-        /*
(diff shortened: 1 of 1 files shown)


savudani8 (approved):

src/parachain/escrow.ts
```diff
@@ -146,29 +146,72 @@ export class DefaultEscrowAPI extends DefaultTransactionAPI implements EscrowAPI
...
-        const [currentBlockNumber, epoch, span, rawSlopeChanges] = await Promise.all([
-            this.systemAPI.getCurrentBlockNumber(),
-            this.api.query.escrow.epoch(),
+        let block;
+        let epoch;
+
+        if (blockNumber) {
+            block = new BN(blockNumber);
+            const maxEpoch = await this.api.query.escrow.epoch();
+            epoch = await this.findBlockEpoch(block, maxEpoch.toBn());
```
savudani8, line 155: The `maxEpoch` query could be moved inside `findBlockEpoch` since that function is already async. I don't think the caller should be aware of how to fetch the maxEpoch

PR #354, opened 2022-01-25
feat: add AuraApi to fetch authorities

gregdhill:

### DIFF
--- a/src/parachain/aura.ts
+++ b/src/parachain/aura.ts
@@ -1,0 +1,23 @@
+import { ApiPromise } from "@polkadot/api";
+import { AccountId } from "@polkadot/types/interfaces";
+import { Vec } from "@polkadot/types";
+
+/**
+ * @category InterBTC Bridge
+ */
+export interface AuraApi {
+    /**
+     * @returns Array of system authorities.
+     */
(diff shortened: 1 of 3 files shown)


savudani8 (commented):
A quick integration test would be very useful

src/parachain/aura.ts
```diff
@@ -0,0 +1,23 @@
...
+     * @returns Array of system authorities.
+     */
+    getAuthorities(): Promise<Array<AccountId>>;
+}
+
+export class DefaultAuraAPI implements AuraApi {
+    constructor(private api: ApiPromise) { }
+
+    async getAuthorities(): Promise<Array<AccountId>> {
+        const head = await this.api.rpc.chain.getFinalizedHead();
```
nud3l, line 19: I think since polkadot-js 7.x will not support this getFinalizedHead query any longer and the authorities don't prompt any user interaction, we don't need to await the finalized head before reporting the number. Worst case, you have a brief inconsistency when displaying.

Mid-term, I'd like to remove all the finalize head awaits where we don't require them to cut the network query time on calls by 50% and reduce API load on the parachain nodes.

nud3l (changes requested):

PR #355, opened 2022-01-25
fix: fixed issuable capacity calculation, 

theodorebugnet:
and extended type to allow index bufgix re. parachain confirmations displayed


src/parachain/vaults.ts
```diff
@@ -541,20 +547,10 @@ export class DefaultVaultsAPI extends DefaultTransactionAPI implements VaultsAPI
...
-                    new Promise<MonetaryAmount<WrappedCurrency, BitcoinUnit>>((resolve) => {
-                        this.tokensAPI.total(currency as Currency<CurrencyUnit>).then(async (nomination) => {
-                            const capacity = await this.calculateCapacity(
-                                nomination as MonetaryAmount<Currency<CollateralUnit>, CollateralUnit>
-                            );
-                            resolve(capacity);
-                        });
-                    })
-            )
+        const perVaultIssuableAmounts = await Promise.all(
```
nud3l, line 550: This query is very inefficient and has degrades with the number of vaults since you need to query each vault from the parachain node and then loop through all the vaults. Instead, use the TotalUserVaultCollateral for all collateral currencies defined here https://github.com/interlay/interbtc/blob/master/crates/vault-registry/src/lib.rs#L633 plus the total issued wrapped tokens at the moment to get the total issuable amount.

nud3l (changes requested):

savudani8:
Other than what Dom said, LGTM

src/parachain/vaults.ts
```diff
@@ -588,20 +587,27 @@ export class DefaultVaultsAPI extends DefaultTransactionAPI implements VaultsAPI
...
-                            );
-                            resolve(capacity);
-                        });
-                    })
-            )
-        );
-        const totalIssuableAmount = perCollateralIssuableAmounts.reduce((acc, v) => acc.add(v));
+        // get [[wrapped, collateral], amount][] map
+        const perCurrencyPairCollateralAmounts = await this.api.query.vaultRegistry.totalUserVaultCollateral.entries();
+        // filter for wrapped === this.wrapped (as only one wrapped currency is handled at a time currently?)
```
savudani8, line 592: That's correct, just one wrapped currency at the moment

src/parachain/vaults.ts
```diff
@@ -588,20 +587,27 @@ export class DefaultVaultsAPI extends DefaultTransactionAPI implements VaultsAPI
...
+                // .has() is true, hence non-null
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                collateralAmount = collateralAmount.add(amounts.get(collateralCurrency)!);
+            }
+            amounts.set(collateralCurrency, collateralAmount);
+            return amounts;
+        }, new Map<Currency<CollateralUnit>, MonetaryAmount<Currency<CollateralUnit>, CollateralUnit>>());
+        // finally convert the CollateralAmount sums to issuable amounts and sum those to get the total
+        const perCollateralCurrencyIssuableAmounts = await Promise.all(
+            [...perCollateralCurrencyCollateralAmounts.values()].map((collateralAmount) => this.calculateCapacity(collateralAmount)));
```
savudani8, line 609: It's great that you preprocess `perWrappedCurrencyCollateralAmounts` into `[wrapped, collateral]` pairs before calling `calculateCapacity`. Otherwise there would have been unnecessary queries for the exchange rate and secure threshold

src/parachain/vaults.ts
```diff
@@ -588,20 +587,27 @@ export class DefaultVaultsAPI extends DefaultTransactionAPI implements VaultsAPI
...
+        const totalIssuableAmount = perCollateralCurrencyIssuableAmounts.reduce((acc, v) => acc.add(v));
         const issuedAmountBtc = await this.getTotalIssuedAmount();
```
savudani8, line 611: This `await` doesn't depend on having any computation done so you can fetch it along with the `perCurrencyPairCollateralAmounts` in a `Promise.all`

savudani8 (approved):

PR #356, opened 2022-01-26
Transaction api refactoring

savudani8:
- Renames `InterBtcAPI` to `BridgeAPI`
- Refactors usage of `TransactionAPI` to an object passed by reference in each API class, instead of a parent class
- Upgrades to interbtc-standalone `1.6.1`


gregdhill:
While we're anyway breaking the API I wonder if the naming should be more conventional camelCase, i.e. `BridgeApi` instead of `BridgeAPI`.

nud3l (commented):
I think instead of renaming it to BridgeApi, it'd be better to rename to InterlayApi. Ultimately, we will have a range of functions that might not just be limited to bridging but also AMM trades or wasm smart contracts. To account for that, I'd name it like that.

But would be good to have @crypto-engineer and @tomjeatt's opinion on this as well as they are the main consumer of this API and I can see that there could be a collision between using InterlayApi on Kintsugi. To me, it would make most sense that way though to keep consistent branding (which ultimately is I think what this change is mostly about).

In a related note, would be good to make sure we have this change on the radar as a ticket that we need to update the UI and potentially index if we merge this breaking change.

Also agree with Greg to stick with proper camel-case naming and calling it BridgeApi or InterlayApi.

README.md
```diff
@@ -25,7 +25,7 @@ The library assumes you have a version of the [Interlay or Kintsugi networks](ht
 ### Creating an API Instance
 
 To use the library, you will first need to create a PolkadotJS `APIPromise` instance,
-and then instantiate a `InterBTCAPI` instance.
+and then instantiate a `BridgeAPI` instance.
 
 ```ts
 import { createInterbtcAPI } from "@interlay/interbtc";
```
nud3l, line 31: Side note: this will need updating then as well.

crypto-engineer:
@nud3l @savudani8 
> I think instead of renaming it to BridgeApi, it'd be better to rename to InterlayApi. Ultimately, we will have a range of functions that might not just be limited to bridging but also AMM trades or wasm smart contracts. To account for that, I'd name it like that.

If you ask me, I'd like us to name it in some neutral way instead of Interlay or Kintsugi. In our interbtc-ui project, Interlay and Kintsugi terms are widely used for Interlay polkadot app and Kintsugi kusama app versions. So if we used `InterlayApi` regardless of whether it's Interlay or Kintsugi version, it could misleading or confusing since they might be wondering if they should use `KintsugiApi` in Kintsugi version. Actually, I have had similar experience due to this naming collision.
To be honest, I'd like us to rename the repo name `interbtc-ui` too since it's also for "kbtc-ui".

crypto-engineer:
@nud3l @savudani8 @gregdhill 
> While we're anyway breaking the API I wonder if the naming should be more conventional camelCase, i.e. `BridgeApi` instead of `BridgeAPI`.

Actually, for this naming, I would like to follow `BridgeAPI` instead of `BridgeApi` since I think "API" appears more correct to me than "Api" from an abbreviation syntax perspective. Similarly "URL" instead of "Url"

This is PascalCase by the way. :)

gregdhill:
> This is PascalCase by the way. :)

Ah yes you are right oops. Anyway I agree with the other point raised by @crypto-engineer that the name should be more generic than Interlay/Kintsugi. I was fine with `InterBtcApi` because interBTC is the shared protocol, but I also like `BridgeApi` FWIW.

tomjeatt:
I agree with Greg and Anton that a neutral name is better, except I don't think they go quite far enough :) `Bridge` ties it too closely to a single function, `Interlay` too closely to a single platform. I'd suggest looking for something that ties it thematically to the company (`DarnApi` anyone?) but not specifically to a single function of application - if it's completely neutral there's no confusion, i.e. why are we calling smart contracts from the bridge api?

nud3l:
Ok, then let's leave it at `InterbtcApi` for now until we find a good name so that we unblock Anton on the UI.

gregdhill (approved):

PR #357, opened 2022-01-27
feat: add frozen and transferable balance to account

nud3l:
Adds support to get the frozen balance and the transferable balance of an account.


nud3l:
> Integration testing this would have been helpful but I can't see an extrinsic to set that (can only set `free` and `reserved`)

It would be possible to test by doing `escrow.vestedTransfer`. That way, balance is frozen.

PR #363, opened 2022-02-02
feat: get vault governance rewards

savudani8:


theodorebugnet (approved):

savudani8:
Merged as tests are passing but hanging. Need to investigate that in another PR

PR #364, opened 2022-02-03
WIP: democracy API

savudani8:


src/parachain/democracy.ts
```diff
@@ -0,0 +1,97 @@
...
+        const tx = this.api.tx.democracy.second(parsedHash, upperBound);
+        await this.transactionAPI.sendLogged(tx, this.api.events.democracy.Passed);
+    }
+
+    async vote(
+        referendumIndex: number,
+        vote: DemocracyVote
+    ): Promise<void> {
+        const tx = this.api.tx.democracy.vote(referendumIndex, vote);
+        await this.transactionAPI.sendLogged(tx, this.api.events.democracy.Passed);
```
savudani8, line 87: The `successEvent` parameter is probably different for each extrinsic. You can check `augment-api-events.ts` for which event sounds intuitive, or take a quick look at the implementation of the extrinsic in the parachain to see which event gets emitted there (you don't need to understand much Rust)

src/parachain/democracy.ts
```diff
@@ -0,0 +1,97 @@
...
+
+import { addHexPrefix } from "../utils/encoding";
+import { TransactionAPI } from "..";
+
+/**
+ * @category InterBTC Bridge
+ */
+export interface DemocracyAPI {
+    /**
+     * @param proposalHash
```
savudani8, line 12: We should describe the parameters wherever possible, to do this I usually check the descriptions in `augment-api-tx.ts`

src/parachain/democracy.ts
savudani8: The democracy pallet is implemented here: https://github.com/interlay/interbtc/blob/master/crates/democracy/src/lib.rs

You're looking for calls to `deposit_event`

PR #365, opened 2022-02-03
Fix required collateral bug

savudani8:
- RPC call to the parachain is not working because of a polkadot-js bug
- Reimplements the logic in TS and tests the function


tomjeatt (approved):
Looks good Dan - thanks for adding me to the PR, nice to understand what the problem was :)

PR #371, opened 2022-02-23
fix: escrow divide by 0 error and big.js dependency

nud3l:


package.json
```diff
@@ -87,6 +87,9 @@
     "typedoc-plugin-markdown": "^3.11.6",
     "typescript": "^4.3.2"
   },
+  "resolutions":{
```
savudani8, line 90: Why is the resolution required, given this is a direct dependency using the same version?

nud3l: without the resolution, somewhere big.js 5.1.0 is still imported

savudani8 (approved):

PR #372, opened 2022-02-24
fix: check for locktime increase in reward estimation function and consider locktime and balance in reward share

nud3l:
This PR updates the reward estimation to account for https://github.com/interlay/interbtc/pull/470. In this change, the reward pool share is calculated via the remaining locktime and the locked balanced (not just the locked balance).
The reward estimation now reflects that.

The minor update removes the finalized block `await` for reward estimation as it's fine to do this on non-finalized blocks and will save us network time.

### DIFF
--- a/src/parachain/escrow.ts
+++ b/src/parachain/escrow.ts
@@ -15,6 +15,7 @@ import {
 } from "../types";
 import { SystemAPI } from "./system";
 import { TransactionAPI } from ".";
+import Big from "big.js";
 
 /**
  * @category BTC Bridge
@@ -78,19 +79,25 @@ export interface EscrowAPI {
         amount: MonetaryAmount<Currency<U>, U>,
     ): Promise<void>;
     /**
-     * @param amount Governance token amount to lock (e.g. KINT or INTR)
+     * @param unlockHeight Block number to lock until
      */
     increaseUnlockHeight(
         unlockHeight: number
     ): Promise<void>;
     /**
      * @param accountId User account ID
-     * @param amountToLock New amount to add to the current stake
+     * @param amountToLock Optional: New amount to add to the current stake
+     * @param unlockHeight Optional: New block number to lock until
      * @returns The estimated reward, as amount and percentage (APY)
+     * @remarks The function takes two optional arguments: the amountToUnlock and unlockHeight
+     * to calculate the estimated increase in rewards
+     * @remarks The estimate rewards in tokens is determined by the `stake` in the governance
+     * system measured by the `amount * unlockHeight`
      */
     getRewardEstimate<U extends GovernanceUnit>(
         accountId: AccountId,
-        amountToLock?: MonetaryAmount<Currency<U>, U>
+        amountToLock?: MonetaryAmount<Currency<U>, U>,
+        unlockHeight?: number
     ): Promise<{
             amount: MonetaryAmount<Currency<U>, U>,
             apy: number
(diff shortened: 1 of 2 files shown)


src/parachain/escrow.ts
```diff
@@ -148,61 +156,94 @@ export class DefaultEscrowAPI implements EscrowAPI {
...
+        // 3. The user increases only their stake
+        // 4. The user extends the locktime and increases the stake
+
+        // Case 1: no change
+        let newUserStake = userStake;
+        let newTotalStake = totalStake;
+        const newStakedBalance = stakedBalance;
+        // Case 2 - 4
+        if (amountToLock || unlockHeight) {
+            // Maybe update the staked balance and the endBlock
```
nud3l, line 175: @savudani8 when you review, I thought we could just do
```ts
newStakedBalance.amount = stakedBalance.amount.add(amountToLock);
```

But this ends up with a type error:
```
src/parachain/escrow.ts:180:64 - error TS2345: Argument of type 'MonetaryAmount<Currency<U>, U> | undefined' is not assignable to parameter of type 'MonetaryAmount<Currency<{ readonly KINT: 12; readonly Planck: 0; } | { readonly INTR: 10; readonly Planck: 0; }>, { readonly KINT: 12; readonly Planck: 0; } | { readonly INTR: 10; readonly Planck: 0; }>'.
  Type 'undefined' is not assignable to type 'MonetaryAmount<Currency<{ readonly KINT: 12; readonly Planck: 0; } | { readonly INTR: 10; readonly Planck: 0; }>, { readonly KINT: 12; readonly Planck: 0; } | { readonly INTR: 10; readonly Planck: 0; }>'.
```

It works now but it's rather lengthy/round-about way of doing this. Let me know if there is an easier way to type cast this.

src/parachain/escrow.ts
```diff
@@ -148,61 +156,94 @@ export class DefaultEscrowAPI implements EscrowAPI {
...
+                this.governanceCurrency as Currency<GovernanceUnit>
+            );
+            newStakedBalance.amount = stakedBalance.amount.add(monetaryAddedStakedBalance);
+            newStakedBalance.endBlock = unlockHeight ? unlockHeight : stakedBalance.endBlock;
+
+            let addedStake = new Big(0);
+            // Case 2: user only plans to extend the lock
+            if ((amountToLock === undefined || amountToLock.isZero()) && unlockHeight) {
+                // Note: unlockHeight - currentBlockNumber should not be <0
+                addedStake.add(unlockHeight - currentBlockNumber);
```
savudani8, line 187: Since `addedStake` is added to `newUserStake`, it would be wrong to use a block number in that addition. I'm not sure of the parachain calculation but based on your addition below the block number is a multiplier. So `newUserStake` has to be scaled by `(newUnlockHeight - currentBlockNumber) / (unlockHeight - currentBlockNumber)` - it's probably better to handle changes to `newUserStake` in each conditional block individually

src/parachain/escrow.ts
savudani8: The error shows up because the conditional doesn't guarantee that `amountToLock` is defined. There is no short solution to defaulting to a zero amount. You could do:
```
const monetaryAddedStakedBalance = amountToLock || newMonetaryAmount(0, this.governanceCurrency as Currency<GovernanceUnit>);
```

savudani8: But the above might still require casting because `amountToLock` is not guaranteed to match the `governanceCurrency`. I think a better solution is to instantiate the entire API with a governanceCurrency generic.

nud3l:
Tackled in #375 instead. Thanks for the helpful review @savudani8 !

PR #373, opened 2022-02-25
refactor: Run tests against parachain

savudani8:
- uses https://github.com/interlay/parachain-launch to set up a local environment with a relay chain and a registered parachain, along with interbtc-specific services
- makes tests general over the currency of the connected chain
- switches to parachain metadata
- once merged, will enable xcm calls


nud3l:
I've made a couple of changes to the initial PR:
- Updated a couple of tests to make sure that they are passing.
- Removed the `finalizedHeader` getter from pretty much all getter functions since we usually do not care (expectation case follows) for most get functions. Where we get information from an unconfirmed block, the parachain layer will make sure that we do not execute into some form of invalid state.
- EXCEPTION: in the cross-chain case (issue, redeem, replace) we have to work with finalized blocks. Since we cannot revert Bitcoin transactions, we can only return Bitcoin addresses that users or Vaults should send transactions to from finalized blocks.
- Similar to the removing of finalized header getter from getter functions, I've also modified the `sendLogged` transaction to take an additional optional argument to only await transactions being included in a block but not yet finalized. In most, cases this is sufficient (e.g., a token transfer on the parachain). By default, this API uses finalized blocks though that, in doubt, the slower but safer choice is used. Again this is only relevant on request issue, request redeem, and request replace since this is the moment that Bitcoin addresses are submitted. On the execute part, it becomes irrelevant since the parachain will ensure correctness.
- The goal of removing the finalized header awaiting is to improve performance of the lib in downstream packages (i.e., the UI) and also to improve test times.
- To improve test times further, some of the integration tests are now executed in parallel. This happens for crates that have no interaction with each other or where state changes from one integration test should not interfere in the integration test of the other.

# Future Improvements

Noting down improvements that we should aim for in the future. Will convert this to a ticket in notion as well:
- Tests should use as little magic numbers as possible. Right now, there are a lot of places where e.g., some amount to issue or redeem is used but if the configuration of the parachain changes slightly, the tests start to fail. If magic numbers should be used they should be declared at the beginning of the test file so that they can be changed (and seen) in a single place. This will make debugging tests easier.
- Assert error messages should be informative of why the assert failed. I saw a couple of `assert.isTrue` used on a comparison of Bigs and Monetary units. While it's fine to use it, the message reported should contain the expected and actual value so that a tester can understand why the test failed and not just that it failed.
- Ideally, all tests should run in isolation. I'd like to use dedicated vaults and user accounts for issue, redeem, replace, ... so that these tests can also be run in parallel. This should speed up test times significantly.
- There should be very few tests that depend on the result of others. Maybe somewhere in the vault when we want to test liquidation due to an exchange drop it will affect other vaults but this should be the exception.
- We need to include an ergonomic way to batch calls using the lib. Considering that e.g. in the setup, you need to await the set confirmation and set exchange rate calls sequentially instead of batching them together decreases testing performance as well.

PR #374, opened 2022-03-03
Theo/vault block rewards

theodorebugnet:
Not sure how to actually test to confirm correctness, but I am reasonably certain in my in my interpretation of the parachain code.

### DIFF
--- a/src/parachain/rewards.ts
+++ b/src/parachain/rewards.ts
@@ -178,6 +178,39 @@ export class DefaultRewardsAPI implements RewardsAPI {
         );
     }
 
+    async getBlockRewards(
+        vaultAccountId: AccountId,
+        collateralCurrencyId: CollateralIdLiteral,
+        rewardCurrencyId: CurrencyIdLiteral,
+        nominatorId?: AccountId, // defaults to vaultAccountId if undefined
+        nonce?: number
+    ): Promise<MonetaryAmount<Currency<CurrencyUnit>, CurrencyUnit>> {
+        // Step 1. Get amount in reward pool for vault ID
+        const [rewardsStake, rewardsRewardPerToken, rewardsRewardTally] = await Promise.all([
+            this.getRewardsPoolStake(collateralCurrencyId, vaultAccountId),
+            this.getRewardsPoolRewardPerToken(collateralCurrencyId),
+            this.getRewardsPoolRewardTally(rewardCurrencyId, collateralCurrencyId, vaultAccountId),
+        ]);
+        const rewardsRawLazyDistribution = computeLazyDistribution(rewardsStake, rewardsRewardPerToken, rewardsRewardTally);
+        // Step 2. Get info from staking pool for vault
+        nominatorId = nominatorId || vaultAccountId;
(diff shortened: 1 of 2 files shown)


savudani8:
@theodorebugnet You can try porting an integration test from the parachain to test this, if one exists there

nud3l (changes requested):
There are a couple of issues with this PR. I haven't checked the logic yet but:
- The `getBlockRewards` function is not declared on the interface and as such the UI will not be able to access it.
- The block rewards will need to be added to the `vault.getAPY` function in the vault API since this is how we get the APY for a vault (see https://github.com/interlay/interbtc-ui/blob/master/src/pages/Vault/index.tsx#L121)

src/parachain/rewards.ts
```diff
@@ -178,6 +178,39 @@ export class DefaultRewardsAPI implements RewardsAPI {
         );
     }
 
+    async getBlockRewards(
+        vaultAccountId: AccountId,
+        collateralCurrencyId: CollateralIdLiteral,
+        rewardCurrencyId: CurrencyIdLiteral,
```
nud3l, line 184: The reward currency is always the governance currency and I think we have a getter for that/define it globally for the lib. Hence, we could remove that parameter.

src/parachain/rewards.ts
```diff
@@ -178,6 +178,39 @@ export class DefaultRewardsAPI implements RewardsAPI {
         );
     }
 
+    async getBlockRewards(
+        vaultAccountId: AccountId,
```
nud3l, line 182: I thought we had settled to use strings for the account id so that the UI does not have to do these weird polkadot type conversions? Please change this to a string.

src/parachain/rewards.ts
```diff
@@ -209,6 +242,23 @@ export class DefaultRewardsAPI implements RewardsAPI {
...
+        nonce?: number
+    ): Promise<Big> {
+        if (nonce === undefined) {
+            nonce = await this.getStakingPoolNonce(collateralCurrencyIdLiteral, vaultAccountId);
+        }
+        const collateralCurrency = currencyIdToMonetaryCurrency(
+            newCurrencyId(this.api, collateralCurrencyIdLiteral)
+        ) as CollateralCurrency;
+        const vaultId = newVaultId(this.api, vaultAccountId.toString(), collateralCurrency, this.wrappedCurrency);
+        const head = await this.api.rpc.chain.getFinalizedHead();
```
nud3l, line 257: Please don't use the finalized head approach. Only use the finalized head on cross-chain interactions (i.e., where a BTC or other cross-chain transaction is required). Otherwise, directly query the endpoint.

src/parachain/rewards.ts
nud3l: I saw we use the accountId also in other places, so I guess fine for now.

src/parachain/rewards.ts
nud3l: Nice to have though since I think this is consistent with the current vault api.

PR #375, opened 2022-03-08
fix: escrow api fixes

nud3l:
- Removes a division by 0 error from the API
- Corrects the escrow reward stake to be a fixed point representation of the amount * unlock_height instead of a monetary amount
- Multiplies the APY by 100 to get a percentage back
- Adds the unlockHeight as parameter for the reward estimate
- Bumps the package version to 1.7.0

### DIFF
--- a/src/parachain/escrow.ts
+++ b/src/parachain/escrow.ts
@@ -162,7 +162,7 @@ export class DefaultEscrowAPI implements EscrowAPI {
 
         // If there is nothing staked in the system or
         // no rewards are paid, the rewards are 0
-        if (newTotalStake.isZero() || blockReward.isZero()) {
+        if (newUserStake.isZero() || blockReward.isZero()) {
             return {
                 amount: newMonetaryAmount(
                     0,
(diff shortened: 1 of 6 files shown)


src/parachain/escrow.ts
```diff
@@ -134,12 +145,29 @@ export class DefaultEscrowAPI implements EscrowAPI {
...
+    async getRewards<U extends GovernanceUnit>(accountId: AccountId): Promise<MonetaryAmount<Currency<U>, U>> {
+        // Step 1. Get amount in reward pool for the account ID
+        const [rewardStake, rewardPerToken, rewardTally] = await Promise.all([
+            this.getEscrowStake(accountId),
+            this.getRewardPerToken(),
+            this.getRewardTally(accountId)
+        ]);
+        // Step 2. Calculate the rewards that can be withdrawn at the moment
+        // Stake[currencyId, accountId] * RewardPerToken[currencyId] - RewardTally[currencyId, accountId]
+        const rewards = rewardStake.mul(rewardPerToken).sub(rewardTally);
```
savudani8, line 157: @nud3l we already have a function for this in the lib, would be good to reuse it
https://github.com/interlay/interbtc-api/blob/09f7be92435eb6c16f5e7184b0bd016afb5c451c/src/utils/currency.ts#L44

PR #377, opened 2022-03-08
fix: ensure positive and correct APY amount

nud3l:
- Ensures that positive block numbers are used to estimate the APY/rewards

### DIFF
--- a/src/parachain/escrow.ts
+++ b/src/parachain/escrow.ts
@@ -96,12 +96,13 @@ export interface EscrowAPI {
     /**
      * @param accountId User account ID
      * @param amountToLock New amount to add to the current stake
+     * @param blockLockTimeExtension Amount of blocks the stake will be locked for
      * @returns The estimated reward, as amount and percentage (APY)
      */
     getRewardEstimate<U extends GovernanceUnit>(
         accountId: AccountId,
         amountToLock?: MonetaryAmount<Currency<U>, U>,
-        unlockHeight?: number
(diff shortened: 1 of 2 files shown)


src/parachain/escrow.ts
```diff
@@ -194,27 +195,30 @@ export class DefaultEscrowAPI implements EscrowAPI {
...
+        if (amountToLock && blockLockTimeExtension) {
             const monetaryAddedStake = newMonetaryAmount(amountToLock.toBig(), this.governanceCurrency as Currency<GovernanceUnit>);
             newStakedBalance.amount = stakedBalance.amount.add(monetaryAddedStake);
-            newStakedBalance.endBlock = unlockHeight;
+            newStakedBalance.endBlock = stakedBalance.endBlock + blockLockTimeExtension;
 
             newLockDuration = newStakedBalance.endBlock - currentBlockNumber;
             const newStake = amountToLock.toBig().mul(newLockDuration);
             newUserStake = userStake.add(newStake);
             newTotalStake = totalStake.add(newStake);
```
savudani8, line 209: I'm not sure I understand this. I thought that extending the lock increases the stake represented by the previously locked amount as well? The lock extension only applies to the newly locked amount in this snippet

savudani8: Greg suggested it would be simpler to compute the vKint amount first and then use `userReward = stake * (totalReward / total_stake)`

savudani8 (approved):

PR #380, opened 2022-03-11
fix: vault reward pool calculations hardcoded fetching the wrapped currency rewards instead of using the passed parameter

theodorebugnet:


savudani8 (approved):
Nice one :)

PR #386, opened 2022-03-21
fix: inactive vaults would show as active

theodorebugnet:
The object `status.asActive` is always defined when `status.isActive` so this would always filter Vaults as active.


savudani8 (approved):
Thanks for finding this!

PR #396, opened 2022-04-21
chore: cleanup dependencies

gregdhill:

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,33 +55,27 @@
     "@interlay/interbtc-types": "1.6.0",
     "@interlay/monetary-js": "0.6.0",
     "@polkadot/api": "7.7.1",
-    "@types/big.js": "6.1.2",
     "big.js": "6.1.1",
     "bitcoin-core": "^3.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -55,45 +55,35 @@
...
     "big.js": "6.1.1",
     "bitcoin-core": "^3.0.0",
     "bitcoinjs-lib": "^5.2.0",
-    "bn.js": "4.12.0",
-    "cross-fetch": "^3.0.6",
-    "isomorphic-fetch": "^3.0.0",
-    "regtest-client": "^0.2.0",
-    "sinon": "^9.0.3",
-    "ts-mock-imports": "^1.3.0"
+    "bn.js": "5.2.0",
```
nud3l, line 61: Updating bn.js to 5.x does have some unwanted consequences/breaks the UI: https://github.com/interlay/interbtc-ui/pull/176 and https://github.com/polkadot-js/api/issues/4024

Might want to test if that patch would then work with the UI.

PR #405, opened 2022-05-12
fix: use the electrs API that will query mempool transactions by OP_RETURN, to ensure unconfirmed transactions are shown on redeem

theodorebugnet:
Use `/scripthash/:hash/txs` (https://github.com/Blockstream/esplora/blob/master/API.md#get-scripthashhashtxs) instead of `/scripthash/:hash/txs/chain`.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.9.0",
+  "version": "1.9.1",
```
bvotteler, line 3: Might need to refresh from master. Currently, the version on master is already 1.9.1

savudani8 (approved):
Nice, this might also speed up this test since the tx is picked up as soon as it's in the mempool (cc @bvotteler): https://github.com/interlay/interbtc-api/blob/0dc9c513639ba10aa54dc0d686517bed0cd99f4c/test/integration/parachain/release/redeem.test.ts#L152

PR #406, opened 2022-05-13
Fix vault replacement request test

bvotteler:
Separated vault replacement tests to individually test success and failure. Unskipped test again.
Other maintenance work.

### DIFF
--- a/test/integration/parachain/staging/sequential/replace.test.ts
+++ b/test/integration/parachain/staging/sequential/replace.test.ts
@@ -2,10 +2,12 @@ import { ApiPromise, Keyring } from "@polkadot/api";
 import { KeyringPair } from "@polkadot/keyring/types";
 import { 
     DefaultInterBtcApi, 
+    DefaultTransactionAPI, 
     getCorrespondingCollateralCurrency, 
     InterBtcApi, 
     InterbtcPrimitivesVaultId, 
-    newMonetaryAmount
+    newMonetaryAmount,
+    tickerToCurrencyIdLiteral
 } from "../../../../../src/index";
 
 import { BitcoinCoreClient } from "../../../../../src/utils/bitcoin-core-client";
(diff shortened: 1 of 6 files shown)


test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -102,7 +119,38 @@ describe("replace", () => {
...
+
+        it("should fail vault replace request if not having enough tokens", async () => {
+            const replaceAmount = dustValue;
+
+            interBtcAPI.setAccount(vault_2);
+
+            // check precondition: vault does not hold enough issued tokens to request a replace
+            const tokensInVault = await interBtcAPI.vaults.getIssuedAmount(
+                newAccountId(api, vault_2.address),
+                tickerToCurrencyIdLiteral(collateralCurrency.ticker)
```
savudani8, line 132: I think you intended to use the wrapped currency here. Something like `tickerToCurrencyIdLiteral(vault_2_id.currencies.wrapped.ticker)`. Unfortunate that `.toBig(wrappedCurrency.base)` didn't catch it, I think it's because the unit value is the same.

bvotteler: Ah, thank you for the catch.
I was stumbling around `vault_2_id` 's fields to find something useful. That's the one I couldn't find.

test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -102,7 +119,38 @@ describe("replace", () => {
...
+            );
+
+            try {
+                await interBtcAPI.replace.request(
+                    replaceAmount, 
+                    currencyIdToMonetaryCurrency(vault_3_id.currencies.collateral) as CollateralCurrency
+                );
+                assert.fail("Expected error to be thrown due to lack of issued tokens for vault, but call completed.");
+            } catch (e) {
+                assert.isTrue(e instanceof Error, "Expected replace request to fail with Error");
```
savudani8, line 150: Once you refactor to the builder pattern, maybe a better way to do it is to `sendLogged` with the last parameter set to the `ExtrinsicFailed` event.

test/integration/parachain/staging/sequential/replace.test.ts
savudani8: Ah, I got confused myself because of the signature. The collateral argument is for the vault id. That's probably another sign to refactor signature to use `VaultId`s everywhere. So your initial implementation was fine, but this is indeed a bit cleaner.

savudani8 (approved):

PR #407, opened 2022-05-16
Fix premium redeem does not filter banned vaults

bvotteler:
Three basic changes:
 - getVaultsWithRedeemableTokens now also includes vaults with status "inactive" as well as "active" (ie. only excludes statuses "liquidated" or "committed theft")
 - getPremiumRedeemVaults now applies the same restrictions as getVaultsWithRedeemableTokens as well as the vault's collateral needing to be below the premium redeem threshold.
 - extracted shared method to check if vault is active or inactive, and has redeemable tokens.

### DIFF
--- a/src/parachain/system.ts
+++ b/src/parachain/system.ts
@@ -25,6 +25,12 @@ export interface SystemAPI {
     subscribeToFinalizedBlockHeads(callback: (blockHeader: Header) => void): Promise<() => void>;
 
     /**
+     * On every new parachain block, call the callback function with the new block header
+     * @param callback Function to be called with every new unfinalized block header
+     */
+     subscribeToCurrentBlockHeads(callback: (blockHeader: Header) => void): Promise<() => void>;
+
+     /**
      * @returns The parachain status code object.
      */
     getStatusCode(): Promise<SecurityStatusCode>;
@@ -55,6 +61,13 @@ export class DefaultSystemAPI {
 
     async subscribeToFinalizedBlockHeads(callback: (blockHeader: Header) => void): Promise<() => void> {
         const unsub = await this.api.rpc.chain.subscribeFinalizedHeads((head) => {
+            callback(head);
+        });
+        return unsub;
+    }
+
+    async subscribeToCurrentBlockHeads(callback: (blockHeader: Header) => void): Promise<() => void> {
+        const unsub = await this.api.rpc.chain.subscribeAllHeads((head) => {
             callback(head);
         });
         return unsub;
--- a/src/parachain/vaults.ts
+++ b/src/parachain/vaults.ts
@@ -763,24 +763,38 @@ export class DefaultVaultsAPI implements VaultsAPI {
         return Promise.reject(new Error("Did not find vault with sufficient locked BTC"));
     }
 
+    private isVaultEligibleForRedeem(vault: VaultExt<BitcoinUnit>, activeBlockNumber: number): boolean {
+        const bannedUntilBlockNumber = vault.bannedUntil || 0;
+        return (
+            (vault.status === VaultStatusExt.Active || vault.status === VaultStatusExt.Inactive) &&
+            vault.issuedTokens.gt(vault.toBeRedeemedTokens) &&
+            bannedUntilBlockNumber < activeBlockNumber
+        );
+    }
+
     async getPremiumRedeemVaults(): Promise<
         Map<InterbtcPrimitivesVaultId, MonetaryAmount<Currency<BitcoinUnit>, BitcoinUnit>>
         > {
         const map: Map<InterbtcPrimitivesVaultId, MonetaryAmount<WrappedCurrency, BitcoinUnit>> = new Map();
-        const vaults = await this.list();
+        const [vaults, activeBlockNumber] = await Promise.all([
+            this.list(),
+            this.systemAPI.getCurrentActiveBlockNumber(),
+        ]);        
+
+        // only non-banned, non-liquidated vaults with liquidity are eligible for redeems
+        const redeemVaults = vaults.filter((vault) => {
+            return this.isVaultEligibleForRedeem(vault, activeBlockNumber);
+        });
+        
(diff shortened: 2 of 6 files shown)


sander2:
Thanks for the fast fix! I don't suppose there is an easy way to have `getPremiumRedeemVaults` be just an additional filter over `getVaultsWithRedeemableTokens`? I.e. to do something like 
```
fn getPremiumRedeemVaults() {
    return getVaultsWithRedeemableTokens.filter((x) => x.isVaultEligibleForRedeem())
}
```

src/parachain/vaults.ts
```diff
@@ -763,24 +763,38 @@ export class DefaultVaultsAPI implements VaultsAPI {
...
+            this.list(),
+            this.systemAPI.getCurrentActiveBlockNumber(),
+        ]);        
+
+        // only non-banned, non-liquidated vaults with liquidity are eligible for redeems
+        const redeemVaults = vaults.filter((vault) => {
+            return this.isVaultEligibleForRedeem(vault, activeBlockNumber);
+        });
+        
         const premiumRedeemVaultPredicates = await Promise.all(
```
theodorebugnet, line 789: I'm honestly not sure why we're using this "predicates" construction, as opposed to just a simple filter. @savudani8 Do you remember by any chance why this was written this way?
If it's not necessary then we could simplify this entire thing even further, down to just a `.filter(vault => return isVaultEligibleForRedeem() && isBelowPremiumRedeemThreshold())`.

theodorebugnet, line 789: I'm honestly not sure why we're using this "predicates" construction, as opposed to just a simple filter. @savudani8 Do you remember by any chance why this was written this way?
If it's not necessary then we could simplify this entire thing even further, down to just a `.filter(vault => return isVaultEligibleForRedeem() && isBelowPremiumRedeemThreshold())`.

bvotteler: As far as I can guess (at least until Dan weighs in): map.prototype.filter is synchronous and cannot deal with async methods / Promises, which `isBelowPremiumThreshold` returns.

src/parachain/vaults.ts
```diff
@@ -763,24 +763,36 @@ export class DefaultVaultsAPI implements VaultsAPI {
...
-                    if (redemableTokens.isZero()) {
-                        resolve(false);
-                    } else {
-                        this.isBelowPremiumThreshold(vault.id).then(resolve).catch(reject);
-                    }
-                });
-            })
+            redeemVaults
+                .map((vault) => this.isBelowPremiumThreshold(vault.id)
+                    });
```
savudani8, line 792: the map doesn't change values in-place AFAIK. You should chain its return value with the other functional operators:
```suggestion
            redeemVaults
                .map((vault) => this.isBelowPremiumThreshold(vault.id)
                .filter((_, index) => premiumRedeemVaultPredicates[index])
                .forEach((vault) => map.set(vault.id, vault.getRedeemableTokens()));
            return map;
                    });
```

src/parachain/vaults.ts
```diff
@@ -763,24 +763,36 @@ export class DefaultVaultsAPI implements VaultsAPI {
...
+        const [vaults, activeBlockNumber] = await Promise.all([
+            this.list(),
+            this.systemAPI.getCurrentActiveBlockNumber(),
+        ]);        
+
+        // only non-banned, non-liquidated vaults with liquidity are eligible for redeems
+        const redeemVaults = vaults.filter((vault) => {
+            return this.isVaultEligibleForRedeem(vault, activeBlockNumber);
+        });
+
```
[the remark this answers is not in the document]
bvotteler: They are almost identical. 

The only difference being that `getVaultsWithRedeemableTokens` returns them in descending order of redeemable tokens, whereas `getPremiumRedeemVaults` did not do that before - which in itself contradicts the `@return` description.

Happy to reference that method, if we are sure the change in the order (until now the list was not ordered) is fine.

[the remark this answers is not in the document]
savudani8: At the moment the ordering is not used AFAIK (@theodorebugnet) and instead the UI generates a random number and uses that as an index to pick a vault. So I think we should introduce a breaking change here, to stop sorting. I'm not aware of any consumers depending on this (the bot wasn't using it either).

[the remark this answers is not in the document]
bvotteler: `getVaultsWithRedeemableTokens` returns a map (vault id to redeemable amount), so I'd need to extract the shared bits of list handling. 
They will have to build the map for themselves, though.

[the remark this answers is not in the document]
savudani8: The line of code linked above converts the map to an array, we can discard that bit too. Very low prio but perhaps expose a utility for converting.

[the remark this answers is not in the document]
bvotteler: I decided to keep ordering out of premium redeem for now. Want to check in with Tom, Anton, and/or Peter to check if they want to add in that bit of randomness in picking a vault id to premium redeem handling on the UI, too (where it does not do that right now)

src/parachain/system.ts
```diff
@@ -60,6 +66,13 @@ export class DefaultSystemAPI {
         return unsub;
     }
 
+    async subscribeToCurrentBlockHeads(callback: (blockHeader: Header) => void): Promise<() => void> {
+        const unsub = await this.api.rpc.chain.subscribeAllHeads((head) => {
+            callback(head);
```
savudani8, line 71: If the callback is async, this will not await. That might not be a problem since the async calls will probably resolve before a new block comes in, but it's something worth considering as it might create race conditions.

bvotteler: Thanks for pointing out. Good catch.
 
Also a consideration for `subscribeToFinalizedBlockHeads` (which I've shamelessly copy/pasta'd for this one)
https://github.com/interlay/interbtc-api/blob/master/src/parachain/system.ts#L56

savudani8 (approved):

PR #408, opened 2022-05-18
Peter/vault toggle issue requests

peterslany:
Adding method to enable or disable issue request with specific vault.

### DIFF
--- a/src/parachain/vaults.ts
+++ b/src/parachain/vaults.ts
@@ -318,10 +318,10 @@ export interface VaultsAPI {
     ): Promise<MonetaryAmount<Currency<GovernanceUnit>, GovernanceUnit>>;
     /**
      * Enables or disables issue requests for given vault
-     * @param vaultAccountId The vault ID whose issuing will be toggled
+     * @param vaultId The vault ID whose issuing will be toggled
      * @param acceptNewIssues Boolean denoting whether issuing should be enabled or not
      */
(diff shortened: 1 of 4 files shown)


savudani8 (commented):
Looks great, thank you! Please also bump the patch version in `package.json`

src/parachain/vaults.ts
```diff
@@ -316,6 +316,12 @@ export interface VaultsAPI {
         vaultCollateralIdLiteral: CollateralIdLiteral,
         governanceCurrencyIdLiteral: GovernanceIdLiteral
     ): Promise<MonetaryAmount<Currency<GovernanceUnit>, GovernanceUnit>>;
+    /**
+     * Enables or disables issue requests for given vault
+     * @param vaultAccountId The vault ID whose issuing will be toggled
+     * @param acceptNewIssues Boolean denoting whether issuing should be enabled or not
+     */
+    toggleIssueRequests(vaultAccountId: InterbtcPrimitivesVaultId, acceptNewIssues: boolean): Promise<void>;
```
savudani8, line 324: the name of the parameter should be `vaultId` rather than `vaultAccountId`. VaultId is the tuple that includes the accountId and the currencyPair

peterslany: Ah, yes, I forgot to change the parameter name in the interface. Thanks!

savudani8 (approved):

PR #409, opened 2022-05-20
Fix test issue and redeem

bvotteler:
"Fix" issue and redeem test by awaiting `assert.isRejected` from previous test.
That test's actions seem to run interference on issue and redeem tests following straight afterward if the promise rejection is not awaited. 
This also fixes the test that runs before issue and redeem: "should fail if no account is set"


savudani8 (approved):
Nice one!

PR #412, opened 2022-05-23
feat: improved hrmp channel opening script

sander2:
Now you only need to modify the consts in the top of the script when using the script for opening new channels. Also you don't have to manually encode relay calls anymore - it's now embedded in this script.

Edit: I also made everything configurable by cli arguments, and added the option to automatically submit a governance proposal with the generated extrinsic.

Example usage: `yarn hrmp-setup --action batched --destination-parachain-id 2085 --refund-hex-address 0xa2e1685f62b2a1a996a2a1ba10ac6836c2b72174cab1bd1c6907454e6365fb70 --submit-proposal`

### DIFF
--- a/scripts/hrmp-setup.ts
+++ b/scripts/hrmp-setup.ts
@@ -11,20 +11,22 @@ import { XcmV1MultiLocation } from "@polkadot/types/lookup";
 import type { 
     BTreeMap, Bytes, Compact, Enum, Null, Option, Result, Struct, Text, U256, U8aFixed, Vec, bool, i128, i32, i64, u128, u16, u32, u64, u8
 } from "@polkadot/types";
+import { SubmittableExtrinsic } from "@polkadot/api/types";
 
-
-const PARACHAIN_ENDPOINT = "ws://127.0.0.1:9988";
-const ACCOUNT_URI = "//Alice";
-
-// const PARACHAIN_ENDPOINT = "wss://api-dev-moonbeam.interlay.io/parachain";
-// const ACCOUNT_URI = "quick sense network ozone ostrich bone hole possible timber clog urban primary//sudo/1";
+// Modify these consts when using this script for new channels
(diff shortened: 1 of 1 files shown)


savudani8 (approved):

scripts/hrmp-setup.ts
```diff
@@ -121,27 +151,64 @@ function printExtrinsic(name: string, extrinsic: SubmittableExtrinsic<"promise">
...
+    printExtrinsic(name, extrinsic, endpoint);
+
+    console.log("Please check the printed extrinsic and enter the seed phrase to submit the proposal.");
+    if (shouldSubmit) {
+        let rl = readline.createInterface({
+            input: process.stdin,
+            output: process.stdout
+        });
+        const it = rl[Symbol.asyncIterator]();
+        const seed = await it.next();
```
savudani8, line 164: I think it would be more secure to have the user set the seed in the environment than pass it through stdin, but if this isn't for prod it's alright.

scripts/hrmp-setup.ts
```diff
@@ -121,27 +151,64 @@ function printExtrinsic(name: string, extrinsic: SubmittableExtrinsic<"promise">
     console.log("");
 }
 
+async function maybeSubmitProposal(name: string, extrinsic: SubmittableExtrinsic<"promise">, endpoint: string, api: ApiPromise, shouldSubmit: boolean) {
+    printExtrinsic(name, extrinsic, endpoint);
+
+    console.log("Please check the printed extrinsic and enter the seed phrase to submit the proposal.");
+    if (shouldSubmit) {
```
savudani8, line 158: can `return;` if `!shouldSubmit` to remove the indentation below

savudani8 (approved):

scripts/hrmp-setup.ts
sander2: Would it? I figured stdin would be more secure since it wouldn't be saved in bash history

PR #414, opened 2022-05-23
fix: reject promise if no issued tokens

tomjeatt:
`getBlockRewardAPY` function returns rejected promise when `globalStake` is zero. 

~@bvotteler could you take a look at this? Don't know if this is the best way to handle it. There's a bug when we have zero issued tokens (at https://github.com/interlay/interbtc-api/blob/master/src/parachain/vaults.ts#L497).~

~Dan mentioned that the fix here is to reject the promise - don't know if this is the correct place to reject it though, or if we should be handling this in the `getBlockRewardAPY` function rather than higher up.~

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.10.3",
+  "version": "1.10.4",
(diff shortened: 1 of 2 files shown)


bvotteler (approved):
LGTM 👍

PR #415, opened 2022-05-26
Add multi currency vaults

bvotteler:
Add KINT as collateral

- Updated tests/code to work with upgraded docker images (parachain version 1.13.1, clients version 1.11.2)
- Added new collateral currency, KINT
- Updated tests to run for KSM as well as KINT collateral currencies
- Enabled and fixed test checking vault theft and liquidation
- Update @polkadot/api to version 8.6.2

Breaking changes to exposed API:
- `VaultsAPI.reportVaultTheft` now takes vault ID (InterbtcPrimitivesVaultId) instead of only account ID
- `VaultsAPI.isVaultFlaggedForTheft` now takes vault ID (InterbtcPrimitivesVaultId) instead of account ID and collateral currency literal ID
- Removed `ConstantsAPI.getTransactionByteFee()`; constant no longer exists


src/types/currency.ts
```diff
@@ -47,13 +48,13 @@ export type CollateralIdLiteral =
     | CurrencyIdLiteral.KINT
     | CurrencyIdLiteral.INTR;
 
-export const CollateralAmount = [PolkadotAmount, KusamaAmount];
+export const CollateralAmount = [PolkadotAmount, KusamaAmount, KintsugiAmount];
```
nud3l, line 51: Suggestion: I think in the mid-term it would be beneficial to get the collateral currencies at runtime from the parachain itself. Otherwise, for any new added collateral asset, the lib needs to be updated instead of us being able to deploy collateral assets without lib/UI upgrades.

bvotteler: A bit late to reply (my apologies for that), but: 100% agreed.
Would be nice if we could generate the types from chain specs.

nud3l: Ok, then let's do that as a follow-up PR?

test/integration/parachain/release/redeem.test.ts
```diff
@@ -86,56 +121,93 @@ describe("redeem", () => {
...
             const amountToSteal = newMonetaryAmount(0.00001, wrappedCurrency, true);
-            const btcTxId = await vaultBitcoinCoreClient.sendToAddress(foreignBitcoinAddress, amountToSteal);
-            // it takes about 15 mins for the theft to be reported
-            await DefaultTransactionAPI.waitForEvent(api, api.events.relay.VaultTheft, 17 * 60000);
+            const {txid: btcTxId} = await vaultBitcoinCoreClient.sendBtcTxAndMine(foreignBitcoinAddress, amountToSteal, 3);
+            // wait for tx inclusion and block finalization.
+            await waitForBlockFinalization(bitcoinCoreClient, userInterBtcAPI.btcRelay);
+
+            // report theft
+            await reporterInterBtcAPI.vaults.reportVaultTheft(vaultToLiquidateId, btcTxId);
```
savudani8, line 133: Nit: Anyone can report theft, so you could have reused `userInterBtcAPI` instead of instantiating `reporterInterBtcAPI`

test/integration/parachain/staging/sequential/refund.test.ts
```diff
@@ -33,42 +50,60 @@ describe("refund", () => {
...
+                true,
+                true
+            );
+            const refund = await interBtcAPI.refund.getRequestByIssueId(issueResult.request.id);
+            assert.notEqual(
+                refund.amountBtc.toString(), 
+                "0", 
+                `Expected non-zero amount for refund request with vault 3 (${currencyTicker})`
+            );
+            const refundRequests = await interBtcAPI.refund.list();
```
savudani8, line 101: The length of this list should be checked after the for-loop ends, with an expected length of at least `vault_3_ids.length`. We can merge without this change though, since the assert on the line above already checks this essentially.

test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -80,70 +81,91 @@ describe("replace", () => {
...
+                    userAccount,
+                    issueAmount,
+                    vault_3_id
+                );
+    
+                interBtcAPI.setAccount(vault_3);
+                await interBtcAPI.replace.request(
+                    replaceAmount, 
+                    currencyIdToMonetaryCurrency(vault_3_id.currencies.collateral) as CollateralCurrency
+                );
```
savudani8, line 100: Would be cleaner to refactor this to use the `callWith` utility.

test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -80,70 +81,91 @@ describe("replace", () => {
...
+    
+                // fetch tokens held by vault
+                const tokensInVault = await interBtcAPI.vaults.getIssuedAmount(
+                    newAccountId(api, vault_2.address),
+                    currencyIdToLiteral(vault_2_id.currencies.collateral)
                 );
-                assert.fail("Expected error to be thrown due to lack of issued tokens for vault, but call completed.");
-            } catch (e) {
-                assert.isTrue(e instanceof Error, "Expected replace request to fail with Error");
+
```
savudani8, line 153: Would be cleaner to refactor this to use the `callWith` utility.

test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -80,70 +81,91 @@ describe("replace", () => {
...
+                // make sure vault does not hold enough issued tokens to request a replace
+                const replaceAmount = dustValue.add(tokensInVault);
+                
+                try {
+                    await interBtcAPI.replace.request(
+                        replaceAmount, 
+                        currencyIdToMonetaryCurrency(vault_2_id.currencies.collateral) as CollateralCurrency
+                    );
+                    assert.fail(`Expected error to be thrown due to lack of issued tokens
+                        for vault (collateral: ${currencyTicker}), but call completed.`);
```
savudani8, line 163: Would be cleaner to check that the Promise is rejected, also not blocking

savudani8:
For later: We should think of a way to make it clear that the tests are running on new parachain and clients versions - maybe tag parachain-launch versions. The yarn `docker-parachain-start` script just pulls the `master` branch of the parachain-launch script, so reviewers of this PR would be unaware of any changes.

savudani8 (approved):
LGTM, all my comments can be fixed later

test/integration/parachain/staging/sequential/refund.test.ts
bvotteler: Good spot, thanks. I have a follow-up task to address this.

test/integration/parachain/staging/sequential/replace.test.ts
bvotteler: Ah, nice! Thank you for that. Added these to the follow up task, too.

test/integration/parachain/staging/setup/initialize.test.ts
```diff
@@ -103,8 +98,7 @@ describe("Initialize parachain state", () => {
         userInterBtcAPI = new DefaultInterBtcApi(api, "regtest", userAccount, ESPLORA_BASE_PATH);
         sudoInterBtcAPI = new DefaultInterBtcApi(api, "regtest", sudoAccount, ESPLORA_BASE_PATH);
 
-        wrappedCurrency = userInterBtcAPI.getWrappedCurrency();
-        collateralCurrency = getCorrespondingCollateralCurrency(userInterBtcAPI.getGovernanceCurrency());
+        collateralCurrency = getCorrespondingCollateralCurrencies(userInterBtcAPI.getGovernanceCurrency())[0];
```
gregdhill, line 101: Does this implicitly assume `KSM`? What happens to the tests if I change the return value of `getCorrespondingCollateralCurrency`?

test/integration/parachain/staging/sequential/replace.test.ts
bvotteler: Also added to the follow-up task.
Thank you for the suggestion.

test/integration/parachain/staging/sequential/vaults.test.ts
```diff
@@ -36,45 +35,43 @@ import {
...
     let vault_to_liquidate: KeyringPair;
     let vault_to_ban: KeyringPair;
     let vault_1: KeyringPair;
-    let vault_1_id: InterbtcPrimitivesVaultId;
+    let vault_1_ids: Array<InterbtcPrimitivesVaultId>;
     let vault_2: KeyringPair;
-    let vault_2_id: InterbtcPrimitivesVaultId;
     let vault_3: KeyringPair;
-    let vault_3_id: InterbtcPrimitivesVaultId;
+    let vault_3_ids: Array<InterbtcPrimitivesVaultId>;
```
gregdhill, line 50: Super minor nit, these should all be camelCase

bvotteler:
> We should think of a way to make it clear that the tests are running on new parachain and clients versions - maybe tag parachain-launch versions. The yarn docker-parachain-start script just pulls the master branch of the parachain-launch script, so reviewers of this PR would be unaware of any changes.

Yes! I was thinking along those lines, too.
Certainly a fan of tagging. Using separate versions for our fork of the official parachain-launch can get tricky. So I want to think about that for a bit.

I also want to at least check if we can possibly go back to the upstream version of parachain-launch rather than maintaining a fork, and keep the configuration details in the lib to then be generated as part of `yarn build`.

test/integration/parachain/staging/fee.test.ts
```diff
@@ -33,33 +41,40 @@ describe("fee", () => {
...
-
-    it("should getGriefingCollateral for replace", async () => {
-        const exchangeRateValue = new Big("3855.23187");
-        const exchangeRate = new ExchangeRate<
-            Bitcoin,
-            BitcoinUnit,
-            typeof collateralCurrency,
-            typeof collateralCurrency.units
+    // TODO: re-enable and use mocking (see getPremiumRedeemVaults test for example)
+    it.skip("should getGriefingCollateral for issue", async () => {
```
gregdhill, line 45: I think we could probably remove these tests but open to opinion. The most important thing for me is that the parsing and currency conversions are correct which could even be done in unit tests (if not already) instead of testing the setter and getter (which is provided by polkadot-js).

test/integration/parachain/staging/sequential/issue.test.ts
```diff
@@ -221,44 +230,46 @@ describe("issue", () => {
...
+                    const requestResult = requestResults[0];
+    
+                    // Wait for issue expiry callback
+                    await new Promise<void>((resolve, _) => {
+                        // userInterBtcAPI.issue.subscribeToIssueExpiry(newAccountId(api, userAccount.address), (requestId) => {
+                        //     if (stripHexPrefix(requestResult.id.toString()) === stripHexPrefix(requestId.toString())) {
+                        //         resolve();
+                        //     }
+                        // });
+                    });
```
gregdhill, line 256: Same here as above, for a future PR?

test/integration/parachain/staging/sequential/issue.test.ts
```diff
@@ -221,44 +230,46 @@ describe("issue", () => {
...
+                    assert.isTrue(issueRequest.status === IssueStatus.Cancelled, "Failed to cancel issue request");
+    
+                    // Set issue period back to its initial value to minimize side effects.
+                    await sudo(userInterBtcAPI, () => userInterBtcAPI.issue.setIssuePeriod(initialIssuePeriod));
+    
+                } catch (e) {
+                    // Set issue period back to its initial value to minimize side effects.
+                    await sudo(userInterBtcAPI, () => userInterBtcAPI.issue.setIssuePeriod(initialIssuePeriod));
+                    throw e;
+                }
```
gregdhill, line 270: Good idea to reset the state, though I wonder if we should have some common routine that runs after most tests to "re-genesis" from a blank state. However I guess that may dramatically slowdown the pipeline if Substrate doesn't have an easy way to do that though.

test/integration/parachain/staging/sequential/oracle.test.ts
```diff
@@ -6,47 +6,53 @@ import Big from "big.js";
...
-        const exchangeRate = await interBtcAPI.oracle.getExchangeRate(collateralCurrency);
-        const expectedCollateral = exchangeRate.toBig(undefined).mul(bitcoinAmount.toBig(BitcoinUnit.BTC)).round(0, 0);
-
-        const collateralAmount = await interBtcAPI.oracle.convertWrappedToCurrency(bitcoinAmount, collateralCurrency);
-        assert.equal(collateralAmount.toBig(collateralCurrency.base).round(0, 0).toString(), expectedCollateral.toString());
+    it("should convert satoshi to collateral currency", async () => {
+        for(const collateralCurrency of collateralCurrencies) {
+            const bitcoinAmount = BitcoinAmount.from.BTC(100);
+            const exchangeRate = await interBtcAPI.oracle.getExchangeRate(collateralCurrency);
+            const expectedCollateral = exchangeRate.toBig(undefined).mul(bitcoinAmount.toBig(BitcoinUnit.BTC)).round(0, 0);
```
gregdhill, line 47: Why `undefined`?

test/integration/parachain/staging/sequential/vaults.test.ts
```diff
@@ -114,123 +111,202 @@ describe("vaultsAPI", () => {
...
+            assert.isAtLeast(premiumRedeemVaults.size, 1);
+
+            // locate the amount for the current vault
+            let premiumRedeemAmount: MonetaryAmount<WrappedCurrency, BitcoinUnit> | undefined = undefined;
+            for (const [vaultId, amount] of premiumRedeemVaults) {
+                if (encodeVaultId(vaultId) === encodeVaultId(vault_3_id)) {
+                    premiumRedeemAmount = amount;
+                    break;
+                }
+            }
```
gregdhill, line 248: There is probably a more functional way to write this logic but that's probably more a nit from my side

test/integration/parachain/staging/sequential/vaults.test.ts
```diff
@@ -114,123 +111,202 @@ describe("vaultsAPI", () => {
...
+    }).timeout(10 * 60000);
 
     it("should getLiquidationCollateralThreshold", async () => {
-        const threshold = await interBtcAPI.vaults.getLiquidationCollateralThreshold(collateralCurrency);
-        assert.equal(threshold.toString(), "1.1");
+        const expectedThresholdByTicker: Map<string, string> = new Map([
+            [Polkadot.ticker, "1.1"],
+            [Kusama.ticker, "1.1"],
+            [Kintsugi.ticker, "2"]
+        ]);
```
gregdhill, line 267: Are these set during initialization? If not we should definitely review our use of magic numbers in tests so us changing the genesis on the parachain doesn't randomly break things here.

gregdhill (commented):
Overall this looks good, though it is hard to comment on whether there is enough to catch flaws in our multi-collateral architecture. There are a fair number of skipped and commented tests but I assume these are for future work? We should definitely sync on what can be moved into the parachain / client tests or even removed to make your life easier. I like the overly verbose nature of the assertions and the reduced reliance on test ordering - idempotent tests make for easier debugging. Anyway, none of my comments are particularly critical to block merging this as-is so feel free address in future PRs.

test/integration/parachain/release/redeem.test.ts
```diff
@@ -86,56 +121,93 @@ describe("redeem", () => {
...
+
+            const finalizedPromise = new Promise<void>((resolve, _) => userInterBtcAPI.system.subscribeToFinalizedBlockHeads(
```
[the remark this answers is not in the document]
bvotteler: Fair point, there are two places where I do that.

test/integration/parachain/release/redeem.test.ts
```diff
@@ -86,56 +121,93 @@ describe("redeem", () => {
...
+                    await issueAndRedeem(
+                        userInterBtcAPI, 
+                        userBitcoinCoreClient, 
+                        userAccount, 
+                        vaultToBanId, 
+                        issueAmount, 
+                        redeemAmount, 
+                        false, 
+                        ExecuteRedeem.False
+                    );
```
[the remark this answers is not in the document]
bvotteler: Oh yes. I hope to have some time to add more builder patterns soon™ :)

test/integration/parachain/release/redeem.test.ts
```diff
@@ -86,56 +121,93 @@ describe("redeem", () => {
...
+                //  wait for head + redeem period + 1
+                // then resolve
+                // Wait for redeem expiry callback
+                await new Promise<void>((resolve, _) => {
+                    // redeemAPI.subscribeToRedeemExpiry(newAccountId(api, userAccount.address), (requestId) => {
+                    //     if (stripHexPrefix(redeemRequest.id.toString()) === stripHexPrefix(requestId.toString())) {
+                    //         resolve();
+                    //     }
+                    // });
+                });
```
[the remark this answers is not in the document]
bvotteler: We had a chat last week with @nud3l to pare down the scope of tests to add/fix to get multi-collateral vaults unblocked sooner rather than later. 
This test fell victim to that.

test/integration/parachain/staging/setup/initialize.test.ts
bvotteler: It only needs one collateral currency, whether that is KSM, or KINT doesn't matter here.
The currency is used in conjunction with vault account ids to wait for them to be available - as means to delay test execution until the parachain / docker images are good to go with vaults registered.

test/integration/parachain/staging/sequential/issue.test.ts
bvotteler: Correct, I just added the looping for the time being.

test/integration/parachain/staging/sequential/issue.test.ts
bvotteler: Once we have instant seal, we might be able to afford re-genesis delays, assuming it can be done.
I'm certainly open to the idea.

test/integration/parachain/staging/sequential/oracle.test.ts
bvotteler: Good catch! That was supposed to be `exchangeRate.toBig()` - looks like me stripping the previously defined, single collateralCurrency, auto-refactored that into `undefined`.
Awkward, but at least not breaking.

test/integration/parachain/staging/sequential/vaults.test.ts
bvotteler: @savudani8 and I were tinkering with that. Turns out, camelCase gets tricky-to-read fast with those numbers thrown in. :)
At least with some of the longer names that we decided to keep snake_case in other tests. So I kept it as is for now anyway.

It feels like there is another angle to refactor here anyway, rather than having a list of `vault_<number>_<id | ids | other_stuff>` variables.

test/integration/parachain/staging/sequential/vaults.test.ts
bvotteler: Right. A version of `array.find().map()` would do the trick.
Keeping this for my clean up task.

test/integration/parachain/staging/sequential/vaults.test.ts
bvotteler: No, those are magic numbers.
Happy to set during initialization - I couldn't find a way when working the tests. 

I think that piece of work would fit in well with the proposed changes by Dom here: https://github.com/interlay/interbtc-api/pull/415#discussion_r883290944

bvotteler:
> There are a fair number of skipped and commented tests but I assume these are for future work? We should definitely sync on what can be moved into the parachain / client tests or even removed to make your life easier.

As for the skipped/commented tests: Yes, they are future work and haven't been deemed critical for the goal of getting multi-collateral vaults out.

As for reducing integration test numbers: Yes, I absolutely would love to get rid of the ones that the parachain and/or clients test themselves.
That, and the introduction of instant-seal should make these tests much, much faster.

test/integration/parachain/release/redeem.test.ts
```diff
@@ -86,56 +121,93 @@ describe("redeem", () => {
...
             const flaggedForTheft = await userInterBtcAPI.vaults.isVaultFlaggedForTheft(
-                vaultToLiquidateId.accountId,
-                currencyIdToLiteral(vaultToLiquidateId.currencies.collateral) as CollateralIdLiteral,
+                vaultToLiquidateId,
                 btcTxId
             );
-            assert.isTrue(flaggedForTheft);
+            assert.isTrue(
+                flaggedForTheft,
+                `Expected vault (collateral: ${collateralCurrency.ticker}) to be flagged for theft, but it was not.`
```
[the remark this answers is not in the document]
bvotteler: Thank you. 
It would drive me bonkers if I saw a failure and wouldn't know which collateral variation failed without manually running the test with breakpoints. :)

test/integration/parachain/staging/fee.test.ts
bvotteler: Good shout, thank you. I'll add that note to the task for addressing this test.

PR #418, opened 2022-06-08
fix: div by zero error for liquidated vault

bvotteler:
Fixing div by zero error for specific case when both nominator and vault have 0 (staked) collateral.
In that event, calculating backingCollateralProportion will return 0.

If however, the nominator has staked collateral, but the vault does not, now will reject with a more meaningful error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.11.1",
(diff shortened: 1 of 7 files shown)


theodorebugnet (approved):
LGTM - might need to look at how exchange rates affect test flakiness but out of scope for this fix

PR #420, opened 2022-06-13
Fix test flakiness caused by oracle

bvotteler:
This fix is to have tests/setup code that modify the exchange rates to use the same values as the oracles do.

Previously, when an oracle would update the price to something other than it was before, it could change the result of getting the current exchange rate (depending on aggregate value) and certain validations, particularly on collaterization rates, would fail seemingly at random as expected values were not met by actual values returned.

### DIFF
--- a/scripts/docker-setup.sh
+++ b/scripts/docker-setup.sh
@@ -2,7 +2,7 @@
 if ! [ -d "local-setup" ]
 then
     git clone https://github.com/interlay/parachain-launch/
-    cd parachain-launch && yarn install
+    cd parachain-launch && git checkout 1.1.0-20220613 && yarn install
     yarn start generate --config=configs/kintsugi.yml --servicesPath=configs/kintsugi-services.yml --yes --output=local-setup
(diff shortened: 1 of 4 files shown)


scripts/docker-setup.sh
```diff
@@ -2,7 +2,7 @@
 if ! [ -d "local-setup" ]
 then
     git clone https://github.com/interlay/parachain-launch/
-    cd parachain-launch && yarn install
+    cd parachain-launch && git checkout 1.1.0-20220613 && yarn install
```
gregdhill, line 5: I feel like this version could get lost easily, I wonder if we should set it from an env config or something?

bvotteler: > I feel like this version could get lost easily, I wonder if we should set it from an env config or something?

I haven't quite figured out how to best do that yet. It feels like, either way, there will be one more place to update the commit hash or tag.

I think my favorite way would be if I could have it in package.json to then be extracted and used in the script... somehow

PR #422, opened 2022-06-16
Test redeem BTC tx fee rate is same as oracle fee rate

bvotteler:
Added new test, and as part of that work upgraded a few dependencies.

Upgraded deps:
- Use latest docker configs from parachain-launch (parachain version 1.14.1, clients version 1.12.1)
- Updated dependencies: 
  - polkadot/api version 8.8.2
  - interbtc-types version 1.7.0


test/integration/parachain/staging/sequential/redeem.test.ts
```diff
@@ -116,6 +118,44 @@ describe("redeem", () => {
...
+    it("should have applied oracle fee rate to redeem transaction", async () => {
+        const oracleBtcFeePerByte = await interBtcAPI.oracle.getBitcoinFees();
+
+        const redeemRequests = await interBtcAPI.redeem.list();
+        for (const redeemRequest of redeemRequests) {
+            const opreturnData = stripHexPrefix(redeemRequest.id.toString());
+            const txId = await interBtcAPI.electrsAPI.getTxIdByOpReturn(opreturnData)
+                .catch((e) => {
+                    // if no opreturn is ready, we may still have to wait for it
+                    return interBtcAPI.electrsAPI.waitForOpreturn(opreturnData, 5 * 60 * 1000, 5000);
```
savudani8, line 130: Since this returns immediately if a tx with the given OP_RETURN is found, should we just `await` this step on each redeem request rather than handling the error?

savudani8: Even better, `waitForOpreturn` returns the result of `getTxIdByOpReturn` so I think the `getTxIdByOpReturn` can be replaced by `waitForOpreturn`.

bvotteler: > Even better, `waitForOpreturn` returns the result of `getTxIdByOpReturn` so I think the `getTxIdByOpReturn` can be replaced by `waitForOpreturn`.

Ah, I did not notice that. Thank you.

And yes, we do need to wait (sometimes) as the tests above just send the redeem request, but do not await for those to be broadcast / confirmed. So in this test, where we look at the output, we may have to wait.

test/integration/parachain/staging/sequential/redeem.test.ts
```diff
@@ -116,6 +118,44 @@ describe("redeem", () => {
...
+                assert.fail(`Could not fetch BTC transaction id for redeem request id ${redeemRequest.id}`);
+                return;
+            }
+
+            // get the actual values on the BTC transaction
+            const btcTx = await interBtcAPI.electrsAPI.getTx(txId);
+
+            const actualTxFeeSatoshi = new Big(btcTx.fee || 0);
+            const actualTxWeight = new Big(btcTx.weight || 0);
+            const actualTxVsize = actualTxWeight.div(4).round(0, RoundingMode.RoundUp);
```
savudani8, line 142: Would be useful to add a comment explaining that Vsize is equal to 4 weight units and that is why we divide. Could even be a util function.

test/integration/parachain/staging/sequential/redeem.test.ts
```diff
@@ -116,6 +118,44 @@ describe("redeem", () => {
...
+            if (actualTxVsize.eq(0)) {
+                assert.fail(`Invalid actual tx vsize of 0, cannot calculate fee rate for redeem request id ${redeemRequest.id}`);
+                return;
+            }
+            
+            const actualFeeRateSatoshiPerByte = actualTxFeeSatoshi.div(actualTxVsize);
+            expect(actualFeeRateSatoshiPerByte.toNumber())
+                .to.be.closeTo(
```
savudani8, line 150: Good to learn about `closeTo()`, nice addition

savudani8 (approved):

PR #423, opened 2022-06-17
fix: enable rpc decoration

sander2:
Without this fix, scripts output `API/INIT: RPC methods not decorated`, and when trying to use rpcs the script crashes with `TypeError: Cannot read property '<function>' of undefined`.

With this fix, I can successfully call rpcs.

I don't know too much about this code base, so I don't know if adding the types and rpc definitions has any other effects besides the intended one of enabling rpcs.

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -22,7 +22,9 @@ export function createProvider(endpoint: string, autoConnect?: number | false | 
 
 export function createSubstrateAPI(endpoint: string, autoConnect?: number | false | undefined): Promise<ApiPromise> {
     const provider = createProvider(endpoint, autoConnect);
-    return ApiPromise.create({ provider });
+    const types = getAPITypes();
+    const rpc = getRPCTypes();
+    return ApiPromise.create({ provider, types, rpc });
 }
 
 export async function createInterBtcApi(


sander2:
> Better to delay until @bvotteler has time to try the auto-generated RPC types.

I don't think the RPCs themselves can be auto-generated - we'll always have to define them manually afaik. Of course, the argument and return types could potentially refer to the auto-generated types. Is this what you mean?

As for the way forward, I think it'd be a good idea to remove all rpc definitions, and only add individual ones back in as needed, testing them when they're added.

bvotteler:
> As for the way forward, I think it'd be a good idea to remove all rpc definitions, and only add individual ones back in as needed, testing them when they're added.

I agree.
 
Also: We will need a fairly strong set of integration tests for those RPC we keep/add in if we cannot rely on the typing telling us of breaking changes.

nud3l:
I'd be in favor of adding the RPCs gradually and replacing some of the lib re-implementations of parachain logic.

From my perspective, for most of the places where we would use RPC calls, we already have the tests in place. Just that these tests are related to lib logic instead of parachain logic. Adding in RPCs for that would have the advantage that we could simplify the tests in the lib by:
- Only verifying that return types are correct
- Move the logic testing into the parachain which would overall improve the runtime of the lib tests

@bvotteler @savudani8 @sander2 I'd suggest revisiting this after iBTC launch and seeing in which order we could move logic from the lib to the parachain. What do you think?

---

As for this PR, I am in favor of merging it since it only enables the RPCs on the substrate level. These RPCs are sometimes quite useful when you are debugging and then I'd kind of expect people to somewhat know what they are doing. Otherwise, you'd have to use a "raw" substrate connection instead of being able to utilize the lib.

bvotteler:
> @bvotteler @savudani8 @sander2 I'd suggest revisiting this after iBTC launch and seeing in which order we could move logic from the lib to the parachain. What do you think?

That approach works for me.

nud3l (approved):

sander2:
Test are failing, but I don't think it's due to the changes in the PR, right @bvotteler ?

bvotteler:
> Test are failing, but I don't think it's due to the changes in the PR, right @bvotteler ?

Hard to say. What has me concerned is that the same test failed twice in a row here, but not in other builds. 

When did you last rebase/merge from master? Might be worth merging master in if it has been a while.

bvotteler:
Closing as we do not need to merge this PR anymore.
PR 442 includes this change: https://github.com/interlay/interbtc-api/pull/442

PR #429, opened 2022-06-28
Add asset registry

bvotteler:
Added initial lib API for the asset registry.

### DIFF
--- a/test/integration/parachain/staging/sequential/asset-registry.test.ts
+++ b/test/integration/parachain/staging/sequential/asset-registry.test.ts
@@ -3,12 +3,13 @@ import { ApiPromise, Keyring } from "@polkadot/api";
 import { KeyringPair } from "@polkadot/keyring/types";
 import { createSubstrateAPI } from "../../../../../src/factory";
 import { ESPLORA_BASE_PATH, PARACHAIN_ENDPOINT, SUDO_URI } from "../../../../config";
-import { DefaultAssetRegistryAPI, DefaultInterBtcApi, DefaultTransactionAPI, getStorageKey, stripHexPrefix } from "../../../../../src";
+import { DefaultAssetRegistryAPI, DefaultInterBtcApi, getStorageKey, stripHexPrefix } from "../../../../../src";
 
 import { StorageKey } from "@polkadot/types";
 import { AnyTuple } from "@polkadot/types/types";
 
 import { OrmlAssetRegistryAssetMetadata } from "@polkadot/types/lookup";
+import { waitForFinalizedEvent } from "../../../../utils/helpers";
 
 describe("AssetRegistry", () => {
     let api: ApiPromise;
@@ -45,16 +46,11 @@ describe("AssetRegistry", () => {
         if (newKeys.length > 0) {
             // clean up assets registered in test(s)
             const deleteKeysInstruction = interBtcAPI.api.tx.system.killStorage(newKeys);
+            await interBtcAPI.api.tx.sudo.sudo(deleteKeysInstruction).signAndSend(sudoAccount);
 
-            const sudoWrappedDeleteInstruction = interBtcAPI.api.tx.sudo.sudo(deleteKeysInstruction);
+            // wait for finalized event
+            await waitForFinalizedEvent(interBtcAPI, api.events.sudo.Sudid);
 
-            await DefaultTransactionAPI.sendLogged(
-                interBtcAPI.api, 
-                sudoAccount, 
-                sudoWrappedDeleteInstruction,
-                api.events.sudo.Sudid,
-                true
-            );
         }
         
         return api.disconnect();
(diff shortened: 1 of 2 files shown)


test/utils/helpers.ts
```diff
@@ -201,4 +211,23 @@ export const bumpFeesForBtcTx = async (
...
+ * @param interBtcApi the api to use for querying
+ * @param event the event to wait for, eg. `api.events.assetRegistry.RegisteredAsset`
+ */
+export const waitForFinalizedEvent = async <T extends AnyTuple>(
+    interBtcApi: InterBtcApi,
+    event: AugmentedEvent<ApiTypes, T>
+): Promise<void> => {
+    return new Promise<void>((resolve, _) => interBtcApi.system.subscribeToFinalizedBlockHeads(
+        async (header) => {
+            const events = await interBtcApi.api.query.system.events.at(header.parentHash);
```
sander2, line 227: Why use the hash of the parent instead of the hash of the finalized block? If there is a reason for that then it'd be good to have a comment stating it

src/parachain/asset-registry.ts
```diff
@@ -0,0 +1,68 @@
...
+
+// shorthand type for the unwieldy tuple
+export type AssetRegistryMetadataTuple = [StorageKey<[u32]>, Option<OrmlAssetRegistryAssetMetadata>];
+ 
+export class DefaultAssetRegistryAPI {
+
+    constructor(private api: ApiPromise) { }
+
+    // not private for easier testing
+    metadataToCurrency(metadata: OrmlAssetRegistryAssetMetadata): Currency<UnitList> {
```
savudani8, line 27: It makes more sense to define this either as a `static` method or as a utility function since it doesn't make use of any `DefaultAssetRegistryAPI` class property.

src/parachain/asset-registry.ts
```diff
@@ -0,0 +1,68 @@
...
+        return entries
+            .filter(([,metadata]) => metadata.isSome)
+            .map(([, metadata]) => metadata.unwrap());
+    }
+
+    async getForeignAssetsAsCurrencies(): Promise<Array<Currency<UnitList>>> {
+        const entries = await this.getAssetRegistryEntries();
+
+        return DefaultAssetRegistryAPI.extractMetadataFromEntries(entries)
+            .map(metadata => this.metadataToCurrency(metadata));
```
savudani8, line 66: nit:
```suggestion
            .map(this.metadataToCurrency);
```

test/integration/parachain/staging/sequential/asset-registry.test.ts
```diff
@@ -0,0 +1,126 @@
...
+            return !previousKeyHashes.includes(newKeyHash);
+        });
+
+        if (newKeys.length > 0) {
+            // clean up assets registered in test(s)
+            const deleteKeysInstruction = interBtcAPI.api.tx.system.killStorage(newKeys);
+            await interBtcAPI.api.tx.sudo.sudo(deleteKeysInstruction).signAndSend(sudoAccount);
+
+            // wait for finalized event
+            await waitForFinalizedEvent(interBtcAPI, api.events.sudo.Sudid);
```
savudani8, line 52: I was about to suggest using `DefaultTransactionAPI.sendLogged(api, sudoAccount, interBtcAPI.api.tx.sudo.sudo(deleteKeysInstruction), api.events.sudo.Sudid)`, but then noticed that for some reason the `successEventType` parameter is [not used ](https://github.com/interlay/interbtc-api/blob/d8446ca0eb272af29212ec646df4531d1e7800da/src/parachain/transaction.ts#L51). Specifying that parameter used to check its existence in the `result.events` [array](https://github.com/interlay/interbtc-api/blob/d8446ca0eb272af29212ec646df4531d1e7800da/src/parachain/transaction.ts#L84). For another time, but we should add that check back.

bvotteler: Yeah, this was a bit of an awkward one. Would be good to at least check if the sudo was successful, but in this case, I hope we end up with dead code as soon as our genesis state contains a few foreign assets - which means I do not need to create them, and therefore, not delete them with those awkward sudos.

test/integration/parachain/staging/sequential/asset-registry.test.ts
```diff
@@ -0,0 +1,126 @@
...
+                // check type
+                assert.equal(
+                    metadata[key]?.constructor.name,
+                    className,
+                    `Expected metadata to have field ${key.toString()} of type ${className}, 
+                    but its type is ${metadata[key]?.constructor.name}.`
+                );
+            }
+        }
+    }).timeout(5 * 60000);
```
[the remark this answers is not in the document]
bvotteler: I'm trying to make this approach a bit easier/faster with the abstractions I've built in. I hope those will evolve over time to make it even more elegant.

As for contributing to typegen... let's just say I don't have the confidence in my own capabilities in that area... yet :)

test/unit/parachain/asset-registry.test.ts
```diff
@@ -0,0 +1,141 @@
...
+            OrmlAssetRegistryAssetMetadata: {
+                name: "Bytes",
+                symbol: "Bytes",
+                decimals: "u32",
+                existentialDeposit: "u128",
+                additional: "InterbtcPrimitivesCustomMetadata"
+            },
+            InterbtcPrimitivesCustomMetadata: {
+                feePerSecond: "u128"
+            }
```
savudani8, line 39: Awesome, `registerTypes` is just what we needed. As a future todo, we should try to fetch the decoded metadata from the running chain and build the expected type dynamically, so the unit tests do break when the parachain is upgraded. Example for `OrmlAssetRegistryAssetMetadata` (the hardcoded `50` array index should be replaced but stil): 
```bash
$ subsee --format json --url http://localhost:9933 | jq ".[1] | .V14.types.types | .[50].type.def"
```

savudani8: I have a feeling that parts of the polkadotjs typegen could also be reused for this, worth checking their tests

test/utils/helpers.ts
```diff
@@ -201,4 +211,31 @@ export const bumpFeesForBtcTx = async (
...
+ * Contrary to @see {@link DefaultTransactionAPI.waitForEvent} which looks at all past events 
+ * and may return due to older events using the same name.
+ * 
+ * Note: This method checks the parent blocks in case the event we are looking for 
+ * was finalized in a block just before this method was called.
+ * 
+ * @param interBtcApi the api to use for querying
+ * @param event the event to wait for, eg. `api.events.assetRegistry.RegisteredAsset`
+ */
+export const waitForFinalizedEvent = async <T extends AnyTuple>(
```
savudani8, line 229: Now that this is refactored, it can replace the duplicated code [here](https://github.com/interlay/interbtc-api/blob/d8446ca0eb272af29212ec646df4531d1e7800da/test/integration/parachain/release/redeem.test.ts#L138) and [here](https://github.com/interlay/interbtc-api/blob/d8446ca0eb272af29212ec646df4531d1e7800da/test/integration/parachain/staging/sequential/replace.test.ts#L102)

test/integration/parachain/staging/sequential/asset-registry.test.ts
savudani8: I noticed you send the tx and then `waitForFinalizedEvent` in some other places too, so it would still be good to replace that by updating the `sendLogged(...)` method

test/utils/helpers.ts
bvotteler: Yep, that's part of my clean up / follow up task for the PR that introduced those bits.

savudani8 (approved):

PR #431, opened 2022-07-04
Fix burn exchange rate calculation

bvotteler:
Fixed burn exchange rate from `x * collateral / (issued + toBeIssued)` to `x * collateral / (issued + toBeIssued - toBeRedeemed)`.

Added unit tests in addition to pre-existing integration tests, particularly for the rejection case.

### DIFF
--- a/test/unit/parachain/redeem.test.ts
+++ b/test/unit/parachain/redeem.test.ts
@@ -1,0 +1,68 @@
+import { expect } from "../../chai";
+import sinon from "sinon";
+import { DefaultRedeemAPI, DefaultVaultsAPI, newMonetaryAmount, VaultsAPI } from "../../../src";
+import { ExchangeRate, KBtc, Kintsugi } from "@interlay/monetary-js";
+
+describe("DefaultRedeemAPI", () => {
+    let redeemApi: DefaultRedeemAPI;
+
+    let stubbedVaultsApi: sinon.SinonStubbedInstance<DefaultVaultsAPI>;
+
+    beforeEach(() => {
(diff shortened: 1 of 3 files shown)


test/unit/parachain/redeem.test.ts
```diff
@@ -0,0 +1,68 @@
...
+                toBeRedeemedTokens: newMonetaryAmount(2, KBtc),
+                collateral: newMonetaryAmount(100, Kintsugi),
+            };
+
+            // stub internal call to reutn our mocked vault
+            stubbedVaultsApi.getLiquidationVault.withArgs(sinon.match.any).resolves(mockVaultExt as any);
+
+            const exchangeRate = await redeemApi.getBurnExchangeRate(Kintsugi);
+            expect(exchangeRate).to.be.an.instanceof(ExchangeRate);
+            expect(exchangeRate.rate.toNumber()).to.be.greaterThan(0);
```
peterslany, line 65: I think it would be nice to have test in here to expect the exact exchange rate to make sure we are returning the correct number to the UI, but I'm not sure how complex would that be. What do you think? :)

peterslany (approved):
Nice job! I left only one small comment about adding one more test, but am not sure about how much effort would that require. Feel free to merge if it's not worth the time :)

test/unit/parachain/redeem.test.ts
bvotteler: While I'm not a fan of magic numbers, it does serve nicely here to detect accidental bugs - like during refactoring.
Thank you for the suggestion.

PR #432, opened 2022-07-04
Tom/vault minimum secure collateral

tomjeatt:
@bvotteler some automatic formatting was applied, presumably as a result of the prettier/eslint changes. I committed these first, so the important changes are in the second commit: https://github.com/interlay/interbtc-api/pull/432/commits/10ef8b40353b22f6c6cb2b3a711bc7702dbd9db9

### DIFF
--- a/src/parachain/vaults.ts
+++ b/src/parachain/vaults.ts
@@ -110,6 +110,12 @@ export interface VaultsAPI {
         collateralCurrency: Currency<C>
     ): Promise<MonetaryAmount<Currency<C>, C>>;
     /**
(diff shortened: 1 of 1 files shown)


bvotteler:
> @bvotteler some automatic formatting was applied, presumably as a result of the prettier/eslint changes. I committed these first, so the important changes are in the second commit: [10ef8b4](https://github.com/interlay/interbtc-api/commit/10ef8b40353b22f6c6cb2b3a711bc7702dbd9db9)

That's to be expected. Thanks for the heads up.

bvotteler (approved):
LGTM

PR #433, opened 2022-07-11
test: add INTR parachain integrations tests

bvotteler:
Run INTR parachain integration tests on push to master (ie. after PR has been closed/accepted).
Run KINT parachain integration tests when opening/pushing to a PR to master.

Previously the integration tests would run using only the KINT parachain on both actions.


scripts/docker-setup.sh
```diff
@@ -1,12 +1,36 @@
 #!/bin/bash
+CHAIN=$(echo $1 | tr '[:lower:]' '[:upper:]')
+
+case $CHAIN in
+  INTR)
+    ;;
+  *)
+    CHAIN=KINT
+    ;;
+esac
```
savudani8, line 10: do these lines do anything?

scripts/setup-parachain-docker.ts
```diff
@@ -1,4 +1,5 @@
 import shell from "shelljs";
+import minimist from "minimist";
```
savudani8, line 2: so far we've been using `yargs`, we should stick to one CLI lib for consistency across repos. Is minimist better? 

Example usage of yargs: https://github.com/interlay/bridge-bot/blob/ed7d42f893520d8fe7d806e12c9d9c0a781a6025/src/index.ts#L19

test/integration/parachain/staging/tokens.test.ts
```diff
@@ -20,22 +21,25 @@ describe("TokensAPI", () => {
...
     let interBtcAPI: InterBtcApi;
+    let collateralCurrencies: Array<Currency<CollateralUnit>>;
 
     before(async () => {
         api = await createSubstrateAPI(PARACHAIN_ENDPOINT);
         const keyring = new Keyring({ type: "sr25519" });
         user1Account = keyring.addFromUri(USER_1_URI);
         user2Account = keyring.addFromUri(USER_2_URI);
         interBtcAPI = new DefaultInterBtcApi(api, "regtest", user1Account, ESPLORA_BASE_PATH);
+        collateralCurrencies =
```
savudani8, line 32: Why run the tests on governance currencies instead of on all of them?

scripts/docker-setup.sh
```diff
@@ -1,12 +1,36 @@
 #!/bin/bash
+CHAIN=$(echo $1 | tr '[:lower:]' '[:upper:]')
+
+case $CHAIN in
+  INTR)
+    ;;
+  *)
+    CHAIN=KINT
```
savudani8, line 8: nit: `echo` something here to instruct the caller that their flag has either been overridden or that a default was applied

PR #434, opened 2022-07-12
Feat: add no vault found response to api

bvotteler:
Added new API to get vault by account id and collateral currency combination.
The new API `VaultsAPI.getOrNull()` returns null if the vault/currency combo does not exist, in contrast to the existing API `VaultsAPI.get()` which rejects if the vault/currency combo cannot be found.

-----
Note: the PR also contains linter fixes (auto fixed) in a separate commit.
The actual change is at commit [68650db](https://github.com/interlay/interbtc-api/pull/434/commits/68650db92f8a5684f3bccb0949d21bf53ad50118)
The linter change is at commit [8ef77d4](https://github.com/interlay/interbtc-api/pull/434/commits/8ef77d41ecbc5c4656e64efa398426bfd7c21a15)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.13.2",
+  "version": "1.13.3",
   "description": "JavaScript library to interact with interBTC and kBTC",
   "main": "build/src/index.js",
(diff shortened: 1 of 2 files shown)


tomjeatt (approved):
💥

PR #435, opened 2022-07-20
Chore: Update monetary-js, and remove currency units

bvotteler:
This PR includes the work to:

- bump monetary version to 0.7.0, 
- remove currency units, and 
- fix compiler/typescript errors

The related work to add foreign assets from the asset registry as collateral currencies will be provided in a stacked PR.

-------
For reference: Notable breaking changes from monetary introduced with 0.7.0
- `MonetaryAmount.toBig()` defaults to returning the base unit amount instead of the atomic unit. ie. for Bitcoin the BTC amount, instead of Satoshi
   To access Satoshi, we can pass in a unit (0 is the "atomic unit", Satoshi in this example.)
- `MonetaryAmount.toString()` now defaults to returning the amount in the base unit instead of the atomic unit. Also, the signature has changed from taking optional units to an optional boolean `atomic` flag. ie. `MonetaryAmount.toString(true)` will return the amount as string in atomic units (Satoshi in the example of Bitcoin.)
- The previous property `MonetaryAmount.zero` has changed to a method, `MonetaryAmount.zero()`

### DIFF
--- a/src/utils/currency.ts
+++ b/src/utils/currency.ts
@@ -2,7 +2,6 @@ import Big, { BigSource } from "big.js";
 import BN from "bn.js";
 import {
     Bitcoin,
-    BitcoinUnit,
     Currency,
     ExchangeRate,
     Interlay,
@@ -14,13 +13,7 @@ import {
     VoteKintsugi,
 } from "@interlay/monetary-js";
 import { InterbtcPrimitivesOracleKey } from "@polkadot/types/lookup";
-import {
-    CurrencyUnit,
-    tickerToCurrencyIdLiteral,
-    GovernanceCurrency,
-    GovernanceUnit,
-    CollateralCurrency,
-} from "../types/currency";
+import { tickerToCurrencyIdLiteral, GovernanceCurrency, CollateralCurrency } from "../types/currency";
 import { ApiPromise } from "@polkadot/api";
 import { FeeEstimationType } from "../types/oracleTypes";
 import { newCurrencyId } from "./encoding";
@@ -30,6 +23,7 @@ Big.PE = 21;
 Big.NE = -12;
 
 export const MBTC_IN_SAT = 100_000;
+export const ATOMIC_UNIT = 0;
 
 export function roundTwoDecimals(input: string): string {
     const number = new Big(input);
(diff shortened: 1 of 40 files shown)


src/utils/currency.ts
```diff
@@ -52,57 +46,51 @@ export function roundLastNDigits(n: number, x: BN | Big | string): string {
...
 }
 
-export function newMonetaryAmount<U extends CurrencyUnit>(
-    amount: BigSource,
-    currency: Currency<U>,
-    base = false
-): MonetaryAmount<Currency<U>, U> {
-    const unit = base ? currency.base : currency.rawBase;
-    return new MonetaryAmount<Currency<U>, U>(currency, amount, unit);
+export function atomicToBaseAmount(atomicAmount: BigSource, currency: Currency): Big {
```
savudani8, line 49: This function should eventually make its way to the monetary lib

src/utils/currency.ts
```diff
@@ -52,57 +46,51 @@ export function roundLastNDigits(n: number, x: BN | Big | string): string {
...
+    return new Big(atomicAmount).div(new Big(10).pow(currency.decimals));
+}
+
+export function newMonetaryAmount(amount: BigSource, currency: Currency, base = false): MonetaryAmount<Currency> {
```
savudani8, line 53: I like that you kept this for backward-compatiblity, even though it's no longer required :) Maybe add a `@deprecated` to inform of the intended usage with the simplified `MonetaryAmount`

src/utils/currency.ts
```diff
@@ -30,6 +23,7 @@ Big.PE = 21;
 Big.NE = -12;
 
 export const MBTC_IN_SAT = 100_000;
+export const ATOMIC_UNIT = 0;
```
savudani8, line 26: Ideally a const exported by the monetary lib?

test/utils/helpers.ts
```diff
@@ -129,8 +122,7 @@ export const vaultStatusToLabel = (status: VaultStatusExt): string => {
 // use the same exchange rate as the running oracles use
 // to avoid flaky tests due to updated prices from the oracle client(s)
 // note: currently the same for all collateral currencies - might change in future
-export const getExchangeRateValueToSetForTesting = <U extends UnitList>(_collateralCurrency: Currency<U>): Big =>
-    new Big("230.0");
+export const getExchangeRateValueToSetForTesting = (_collateralCurrency: Currency): Big => new Big("230.0");
```
savudani8, line 125: I think `_collateralCurrency` was only passed for type inference and can now be removed

savudani8 (approved):
Amazing to be able to simplify the types so much!

test/utils/helpers.ts
bvotteler: I introduced it back in the day in case we need/want different rates for different currencies.

PR #436, opened 2022-07-22
Feat: Add foreign asset model as valid currency to lib

bvotteler:
Add foreign assets to the model to be available in the same ways as monetary-js's Currency was before.

Changes:
1) Add `ForeignAsset`s as Currency, expanding the current Currency model.
Collateral currencies in particular can now be of type `CurrencyExt = Currency | ForeignAsset`, where `ForeignAsset = Currency & { id: number }`.
2) Remove `CurrencyIdLiteral`. Currency tickers are no longer the only primary identifier. ForeignAssets are identified by their respective on-chain ids (numeric) and may theoretically have clashing tickers.
3) Some minor refactoring/moving of methods to more suitable locations (eg. from `types/currency.ts` to `util/currency.ts`)

The best starting point to understand where all the changes come from is to start with what changed in `src/types/currency.ts` [here](https://github.com/interlay/interbtc-api/pull/436/files#diff-6cc7d53fe1f7dc8cf1e80369eb2ba258f456c8bee9aae4323086db7a7d968947)

Breaking changes to most APIs that previously took a `CurrencyIdLiteral`. Those will now take a `CurrencyExt` instead.

### DIFF
--- a/.github/workflows/integration-tests-kint.yml
+++ b/.github/workflows/integration-tests-kint.yml
@@ -4,6 +4,7 @@ on:
     pull_request:
         branches:
             - master
+            - brendon/feat-add-foreign-asset-model
 
 concurrency:
     group: ${{ github.ref }}_kint
--- a/src/utils/encoding.ts
+++ b/src/utils/encoding.ts
@@ -8,6 +8,7 @@ import { Codec } from "@polkadot/types/types";
 import { Moment } from "@polkadot/types/interfaces";
 import { Option } from "@polkadot/types/codec";
 import { Bytes } from "@polkadot/types-codec";
+import { u32 } from "@polkadot/types";
 import {
     InterbtcPrimitivesRedeemRedeemRequest,
     InterbtcPrimitivesReplaceReplaceRequest,
@@ -21,22 +22,13 @@ import {
     InterbtcPrimitivesCurrencyId,
 } from "@polkadot/types/lookup";
 
-import { encodeBtcAddress, FIXEDI128_SCALING_FACTOR } from ".";
+import { currencyIdToMonetaryCurrency, encodeBtcAddress, FIXEDI128_SCALING_FACTOR, isForeignAsset } from ".";
 import { WalletExt, SystemVaultExt } from "../types/vault";
 import { Issue, IssueStatus, Redeem, RedeemStatus, RefundRequestExt, ReplaceRequestExt } from "../types/requestTypes";
 import { SignedFixedPoint, UnsignedFixedPoint } from "../interfaces";
-import {
-    CollateralCurrency,
-    CollateralIdLiteral,
-    CurrencyIdLiteral,
-    currencyIdToLiteral,
-    currencyIdToMonetaryCurrency,
-    tickerToCurrencyIdLiteral,
-    WrappedCurrency,
-    WrappedIdLiteral,
-} from "../types";
+import { CollateralCurrencyExt, CurrencyExt, WrappedCurrency } from "../types";
 import { newMonetaryAmount } from "../utils";
-import { VaultsAPI } from "../parachain";
+import { AssetRegistryAPI, VaultsAPI } from "../parachain";
 
 /**
  * Converts endianness of a Uint8Array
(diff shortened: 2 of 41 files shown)


.github/workflows/integration-tests-kint.yml
```diff
@@ -4,6 +4,7 @@ on:
     pull_request:
         branches:
             - master
+            - brendon/feat-add-foreign-asset-model
```
bvotteler, line 7: I'll remove this in a quick throwaway PR once this work has been merged successfully.

src/utils/encoding.ts
```diff
@@ -158,19 +150,20 @@ export function parseWallet(wallet: VaultRegistryWallet, network: Network): Wall
...
     return {
         toBeIssuedTokens: newMonetaryAmount(vault.toBeIssuedTokens.toString(), wrappedCurrency),
         issuedTokens: newMonetaryAmount(vault.issuedTokens.toString(), wrappedCurrency),
         toBeRedeemedTokens: newMonetaryAmount(vault.toBeRedeemedTokens.toString(), wrappedCurrency),
         collateral: newMonetaryAmount(vault.collateral.toString(), collateralCurrency),
         currencyPair: {
-            collateralCurrency: currencyIdToMonetaryCurrency(vault.currencyPair.collateral),
-            wrappedCurrency: currencyIdToMonetaryCurrency(vault.currencyPair.wrapped),
+            collateralCurrency: await currencyIdToMonetaryCurrency(assetRegistryApi, vault.currencyPair.collateral),
+            wrappedCurrency: await currencyIdToMonetaryCurrency(assetRegistryApi, vault.currencyPair.wrapped),
```
gregdhill, line 166: Minor nit: probably could await both promises concurrently

test/utils/helpers.ts
```diff
@@ -196,9 +198,81 @@ export const waitForFinalizedEvent = async <T extends AnyTuple>(
...
+              })
+            : Promise.resolve();
+
+    const waitForEventPromise = finalized
+        ? waitForFinalizedEvent(interBtcApi, event)
+        : waitForIncludedEvent(interBtcApi, event);
+
+    await Promise.race([waitForEventPromise, timeoutPromise]).then((_) => {
+        clearTimeout(timeoutHandle);
+    });
```
gregdhill, line 275: Ooo this is good to know, `race` presumably returns the first `Promise` to resolve?

test/unit/parachain/asset-registry.test.ts
```diff
@@ -5,11 +5,17 @@ import { StorageKey, u32 } from "@polkadot/types";
...
+import * as allThingsEncoding from "../../../src/utils/encoding";
 
 describe("DefaultAssetRegistryAPI", () => {
     let api: ApiPromise;
     let assetRegistryApi: DefaultAssetRegistryAPI;
     let mockMetadata: OrmlAssetRegistryAssetMetadata;
+    let mockStorageKey: StorageKey<[u32]>;
+    // scraped from integration tests
+    const scrapedStorageKey =
+        "0x6e9a9b71050cd23f2d7d1b72e8c1a625b5f3822e35ca2f31ce3526eab1363fd25153cb1f00942ff401000000";
```
gregdhill, line 17: I think we can construct the expected value instead of having this magic value, there is some info of how to do this in the "Storage Keys" section of this blog: https://www.shawntabrizi.com/substrate/querying-substrate-storage-via-rpc/

gregdhill (commented):
This is a good first step toward drastically reducing the complexity of currency handling. Happy to merge this as-is, no blockers on my end - see review comments for minor suggestions.

gregdhill (approved):

src/utils/encoding.ts
bvotteler: > Minor nit: probably could await both promises concurrently

Good shout, will add to follow-up task. Thank you.

test/unit/parachain/asset-registry.test.ts
bvotteler: > I think we can construct the expected value instead of having this magic value, there is some info of how to do this in the "Storage Keys" section of this blog: https://www.shawntabrizi.com/substrate/querying-substrate-storage-via-rpc/

Thanks for that: yes! 
I want to review this at some point anyway. I think with the mocking I have added later, this isn't really needed anymore and can use dummy values as long as they don't break the format.

PR #437, opened 2022-07-22
Peter/feat script for setting state

peterslany:
Adding script for setting state of parachain by overwriting the storage.
Usage: `yarn set [command] [options]`, available commands are:
```
balance             Sets token balance of an account
liquidationVault    Sets system liquidation value
vaultIssuedTokens   Sets issued tokens of a vault
vaultPremiumRedeem  Sets collateral ratio of vault to be within
                                   premium redeem range
vaultTheft          Sets vault status to theft
vaultBan            Bans vault for next 1000 blocks
vaultUnban          Unbans vault
vaultFund           Funds vault account with collateral token
vaultReward         Sets claimable amount of rewards for vault
```
Passing the `--help` shows helping message to display options for each command. 

Default vault account is obtained from `VAULT_1_URI` env variable. Default vault is then vault consisting of this account and currencies: collateral currency - native relay chain currency, wrapped currency - wrapped currency obtained from parachain constants. 
It is also possible to specify vault by passing options: 
`-a [accountId -c [collateralCurrencySymbol] -w [wrappedCurrencySymbol]`

Example: `yarn set vaultTheft -a 5CiPPseXPECbkjWCa6MnjNokrgYjMqmKndv2rSnekmSK2DjL -c KSM -w KBTC`


src/utils/set-state.ts
```diff
@@ -0,0 +1,549 @@
...
+
+const setVaultUnban = ({ api }: DefaultInterBtcApi) =>
+    disconnectApiOnExit(
+        api,
+        async ({ accountId, collateralSymbol, wrappedSymbol }: SetVaultParamsBase) => {
+            const vaultId = constructVaultId(api, accountId, collateralSymbol, wrappedSymbol);
+
+            const newBannedUntil = null;
+
+            const modifier = (vaultData: MutableVaultData): MutableVaultData => ({ ...vaultData, bannedUntil: <any>newBannedUntil });
```
bvotteler, line 280: Not sure if we can get rid of the warning here (Unexpected any), but if not, please add ignore line above.

bvotteler (approved):
LGTM

src/utils/set-state.ts
```diff
@@ -0,0 +1,550 @@
...
+            const scalingFactor = new Big(Math.pow(10, FIXEDI128_SCALING_FACTOR));
+            const newCollateralAmount = newIssuedTokensInCollateralAmount.toBig().mul(newCollateralizationRatio).mul(scalingFactor);
+            const totalStakeDataHex = api.createType("u128", newCollateralAmount.toFixed()).toHex(true);
+
+            const nonceHex = api.createType("u32", 0).toHex();     // 0 nonce by default
+            const totalStakeStorageKey = getStorageMapItemKey("VaultStaking", "TotalCurrentStake", nonceHex, vaultId.toHex());
+
+            console.log("Setting backing collateral...");
+            await setStorageAtKey(api, totalStakeStorageKey, totalStakeDataHex);
+            console.log(
```
savudani8, line 172: If these setters are meant to do more than visual checking of the UI's layout given a state (I think it should), then this PR should also test that the outcomes are what you expect.

For instance, I believe these would all be tests in the UI and we must be confident they don't break because of the lib.
- `setPremiumRedeem`: test that there is at least on vault that premium redeems can be executed with, and redeem with it
- `setLiquidationVault`: test that burn redeeming works
- `setVaultReward`: test that the vault can actually claim the reward
- the other setters too

src/utils/set-state.ts
```diff
@@ -0,0 +1,550 @@
...
+const setVaultUnban = ({ api }: DefaultInterBtcApi) =>
+    disconnectApiOnExit(
+        api,
+        async ({ accountId, collateralSymbol, wrappedSymbol }: SetVaultParamsBase) => {
+            const vaultId = constructVaultId(api, accountId, collateralSymbol, wrappedSymbol);
+
+            const newBannedUntil = null;
+
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const modifier = (vaultData: MutableVaultData): MutableVaultData => ({ ...vaultData, bannedUntil: <any>newBannedUntil });
```
savudani8, line 281: You can also set `bannedUntil` to a block in the past to unban the vault

src/utils/set-state.ts
```diff
@@ -0,0 +1,550 @@
...
+                    address: {
+                        alias: "a",
+                        type: "string",
+                        demandOption: false,
+                        describe: "address of which to set balance"
+                    }
+                },
+                setBalance(sudoAccountInterBtcApi)
+            )
+            .command(
```
savudani8, line 431: These commands tend to get large, it's fine for now but as a future improvement they can be refactored into their own separate JS modules: https://github.com/yargs/yargs/blob/main/docs/advanced.md#providing-a-command-module

src/utils/set-state.ts
peterslany: @savudani8 do you mean to extend the commands to call the same api methods that are used in the UI or to directly test the UI components? 

Also, I have a task in this sprint to start with the UI test automation, but we will be mocking the api calls, so there won't be need to modify the parachain state (and therefore use these utils).

src/utils/set-state/set-state.ts
```diff
@@ -0,0 +1,564 @@
...
+                },
+                setVaultIssuedTokens(sudoAccountInterBtcApi)
+            )
+            .command(
+                "vaultPremiumRedeem",
+                "Sets collateral ratio of vault to be within premium redeem range",
+                SET_VAULT_PARAMS_BASE,
+                setPremiumRedeem(sudoAccountInterBtcApi)
+            )
+            .command(
```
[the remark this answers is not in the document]
peterslany: Thanks for the heads up, I removed the util.

savudani8 (approved):
Nice work!

PR #439, opened 2022-07-28
Feat: Add aUSD tests

bvotteler:
Add tests for a foreign asset called "aUSD".

This includes registering vaults, and setting thresholds and collateral ceiling as part of test setup.

-----
Added aUSD collateral vaults for the following tests.

Redeem:
- should issue and request redeem

vaultsAPI:
- should get issuable
- should get the required collateral for the vault
- should deposit and withdraw collateral
- should getPremiumRedeemVaults after a price crash
- should getLiquidationCollateralThreshold
- should getPremiumRedeemThreshold
- should get the issuable InterBtc for a vault
- should getFees (prepared; currently skipped)
- should getAPY

### DIFF
--- a/test/integration/parachain/staging/sequential/asset-registry.test.ts
+++ b/test/integration/parachain/staging/sequential/asset-registry.test.ts
@@ -10,9 +10,10 @@ import { StorageKey } from "@polkadot/types";
 import { AnyTuple } from "@polkadot/types/types";
 import { AssetId } from "@polkadot/types/interfaces/runtime";
 import { OrmlTraitsAssetRegistryAssetMetadata } from "@polkadot/types/lookup";
-import { waitForIncludedEvent } from "../../../../utils/helpers";
+import { waitForEvent } from "../../../../utils/helpers";
 
 describe("AssetRegistry", () => {
+    const approx10Blocks = 10 * 12 * 1000;
     let api: ApiPromise;
     let interBtcAPI: DefaultInterBtcApi;
 
@@ -48,10 +49,10 @@ describe("AssetRegistry", () => {
         if (newKeys.length > 0) {
             // clean up assets registered in test(s)
             const deleteKeysInstruction = interBtcAPI.api.tx.system.killStorage(newKeys);
-            await interBtcAPI.api.tx.sudo.sudo(deleteKeysInstruction).signAndSend(sudoAccount);
-
-            // wait for finalized event
-            await waitForIncludedEvent(interBtcAPI, api.events.sudo.Sudid);
+            await Promise.all([
+                waitForEvent(interBtcAPI, api.events.sudo.Sudid, false, approx10Blocks),
+                interBtcAPI.api.tx.sudo.sudo(deleteKeysInstruction).signAndSend(sudoAccount),
+            ]);
         }
 
         return api.disconnect();
@@ -82,9 +83,15 @@ describe("AssetRegistry", () => {
             );
 
             // need sudo to add new foreign asset
-            await interBtcAPI.api.tx.sudo.sudo(callToRegister).signAndSend(sudoAccount);
+            const [eventFound] = await Promise.all([
+                waitForEvent(interBtcAPI, api.events.sudo.Sudid, false, approx10Blocks),
+                interBtcAPI.api.tx.sudo.sudo(callToRegister).signAndSend(sudoAccount),
+            ]);
 
-            await waitForIncludedEvent(interBtcAPI, api.events.sudo.Sudid);
+            assert.isTrue(
+                eventFound,
+                `Sudo event to create new foreign asset not found - timed out after ${approx10Blocks} ms`
+            );
         }
 
         // get the metadata for the asset we just registered
--- a/test/integration/parachain/staging/sequential/asset-registry.test.ts
+++ b/test/integration/parachain/staging/sequential/asset-registry.test.ts
@@ -10,10 +10,10 @@ import { StorageKey } from "@polkadot/types";
 import { AnyTuple } from "@polkadot/types/types";
 import { AssetId } from "@polkadot/types/interfaces/runtime";
 import { OrmlTraitsAssetRegistryAssetMetadata } from "@polkadot/types/lookup";
-import { waitForEvent } from "../../../../utils/helpers";
+import { APPROX_BLOCK_TIME_MS, waitForEvent } from "../../../../utils/helpers";
 
 describe("AssetRegistry", () => {
-    const approx10Blocks = 10 * 12 * 1000;
+    const approx10Blocks = 10 * APPROX_BLOCK_TIME_MS;
     let api: ApiPromise;
     let interBtcAPI: DefaultInterBtcApi;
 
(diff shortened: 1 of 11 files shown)


test/integration/parachain/staging/sequential/asset-registry.test.ts
```diff
@@ -48,10 +49,10 @@ describe("AssetRegistry", () => {
         if (newKeys.length > 0) {
             // clean up assets registered in test(s)
             const deleteKeysInstruction = interBtcAPI.api.tx.system.killStorage(newKeys);
-            await interBtcAPI.api.tx.sudo.sudo(deleteKeysInstruction).signAndSend(sudoAccount);
-
-            // wait for finalized event
-            await waitForIncludedEvent(interBtcAPI, api.events.sudo.Sudid);
+            await Promise.all([
+                waitForEvent(interBtcAPI, api.events.sudo.Sudid, false, approx10Blocks),
+                interBtcAPI.api.tx.sudo.sudo(deleteKeysInstruction).signAndSend(sudoAccount),
```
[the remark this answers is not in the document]
bvotteler: > This usage of polkadot-js can be avoided by doing [something like this](https://github.com/interlay/interbtc-api/blob/ad8c3035a6ae36865399993d7c2eeefdb7cff214/test/integration/parachain/release/redeem.test.ts#L200). Doing this will also print all the events of this extrinsic to stdout.
Ah, good catch, thank you.
I believe I had that left over from a previous (failed) attempt to batch the delete calls.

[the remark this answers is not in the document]
bvotteler: Tried to change it, but the signature of `sudo(InterBtcAPI, () =>  Promise<void>): Promise<void>` can not take `() => SubmittableExtrinsic<"prmoise", ISubmittableResult>` as second parameter.

test/integration/parachain/staging/setup/initialize.test.ts
```diff
@@ -187,4 +279,138 @@ describe("Initialize parachain state", () => {
...
+    it("should set collateral ceiling and thresholds for aUSD", async function () {
+        // only get aUSD if it hasn't been set yet (e.g. when running this test in isolation)
+        !aUSD ? (aUSD = await getAUSDForeignAsset(sudoInterBtcAPI.assetRegistry)) : undefined;
+
+        if (aUSD === undefined) {
+            // no point in completing this if aUSD is not registered
+            this.skip();
+        }
+
+        const sudoThis = async (extrinsic: SubmittableExtrinsic<"promise", ISubmittableResult>) =>
```
savudani8, line 292: There already is a sudo utility [here](https://github.com/interlay/interbtc-api/blob/ad8c3035a6ae36865399993d7c2eeefdb7cff214/test/utils/helpers.ts#L75)

bvotteler: Sadly not for `SubmittableExtrinsic<"promise", ISubmittableResult>` - only for `Promise<void>`.

So, in order to use that, I'd have to create a method returning `Promise<void>` which seems like more overhead than having this little helper.

test/integration/parachain/staging/setup/initialize.test.ts
```diff
@@ -187,4 +279,138 @@ describe("Initialize parachain state", () => {
...
+        const existingCollCcy = getCorrespondingCollateralCurrencies(sudoInterBtcAPI.getGovernanceCurrency())[0];
+        const existingCcyPair = newVaultCurrencyPair(api, existingCollCcy, sudoInterBtcAPI.getWrappedCurrency());
+        // borrow values from existing currency pair
+        const [optionCeilValue, existingSecureThreshold, existingPremiumThreshold, existingLiquidationThreshold] =
+            await Promise.all([
+                sudoInterBtcAPI.api.query.vaultRegistry.systemCollateralCeiling(existingCcyPair),
+                sudoInterBtcAPI.vaults.getSecureCollateralThreshold(existingCollCcy),
+                sudoInterBtcAPI.vaults.getPremiumRedeemThreshold(existingCollCcy),
+                sudoInterBtcAPI.vaults.getLiquidationCollateralThreshold(existingCollCcy),
+            ]);
```
savudani8, line 305: Why not just set some arbitrary values?

bvotteler: > Why not just set some arbitrary values?

No good reason. I just wanted some values that have a "realistic spread" between the thresholds.

test/integration/parachain/staging/setup/initialize.test.ts
```diff
@@ -187,4 +279,138 @@ describe("Initialize parachain state", () => {
...
+
+        // register the vaults with aUSD collateral
+        for (const [vaultKeyringPair, vaultAccountId] of vaultAccountIdAndKeyrings) {
+            const balanceTimeout = 5 * APPROX_BLOCK_TIME_MS;
+            // set balance and wait for event
+            const [balanceEventFound] = await Promise.all([
+                waitForEvent(sudoInterBtcAPI, api.events.tokens.BalanceSet, false, balanceTimeout),
+                sudo(sudoInterBtcAPI, async () => sudoInterBtcAPI.tokens.setBalance(vaultAccountId, freeBalanceToSet)),
+            ]);
+            await sudo(sudoInterBtcAPI, async () =>
```
savudani8, line 389: Is there any reason you're sudo-setting this balance twice?

bvotteler: > Is there any reason you're sudo-setting this balance twice?

Missed that one after some refactoring. Thanks for spotting.

test/integration/parachain/staging/sequential/redeem.test.ts
```diff
@@ -99,7 +99,18 @@ describe("redeem", () => {
...
-        for (const [vault_1_id, vault_2_id] of collateralTickerToVaultIdsMap.values()) {
+        // "usual" scope with hard coded collateral currency (or currencies)
+        const vaultsInScope = Array.from(collateralTickerToVaultIdsMap.values());
+
+        // add a pair of aUSD vaults to the list if aUSD has been registered
+        const aUSD = await getAUSDForeignAsset(interBtcAPI.assetRegistry);
+        if (aUSD !== undefined) {
+            const vault_1_id_ausd = newVaultId(api, vault_1.address, aUSD, wrappedCurrency);
+            const vault_2_id_ausd = newVaultId(api, vault_2.address, aUSD, wrappedCurrency);
+            vaultsInScope.push([vault_1_id_ausd, vault_2_id_ausd]);
```
savudani8, line 110: As I understand it, you're not running an actual aUSD vault binary, and instead just registering one. If so, why does `issueAndRedeem` succeed?

test/integration/parachain/staging/sequential/asset-registry.test.ts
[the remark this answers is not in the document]
savudani8: I see, the current `sudo` utility only works for txs already added to the lib interface, unlike `batch` for instance.

savudani8 (approved):

test/integration/parachain/staging/sequential/redeem.test.ts
savudani8: This is correct vault client behaviour so resolving

PR #440, opened 2022-07-29
feat(escrow): add currency rewards param

danielsimao:


src/types/currency.ts
```diff
@@ -53,6 +53,9 @@ export type GovernanceCurrency = typeof GovernanceCurrency[number];
 export const VotingCurrency = [VoteInterlay, VoteKintsugi];
 export type VotingCurrency = typeof VotingCurrency[number];
 
+export const RewardCurrency = [...GovernanceCurrency, KBtc];
```
savudani8, line 56: This can include all of `WrappedCurrency` since `iBtc` is also going live soon

src/utils/rewards.ts
```diff
@@ -1,22 +1,22 @@
-import { GovernanceCurrency, StakedBalance } from "../types";
+import { RewardCurrency, StakedBalance } from "../types";
 import Big from "big.js";
 import { MonetaryAmount } from "@interlay/monetary-js";
 import { ATOMIC_UNIT, newMonetaryAmount } from "./currency";
 
 // TODO: simplify this, perhaps use builder?
 export function estimateReward(
```
savudani8, line 7: I just realised that this is only used for staking rewards which will always be in governance currency only, so this change should be reverted. Sorry about this!

PR #441, opened 2022-08-01
refactor: pull 1.18.0 which removes theft reporting

gregdhill:


bvotteler (approved):
LGTM

Only minor clean up left over (not blocking): remove unused arg in vaults.ts `parseVault(...)` method.
We can add that to the backlog

PR #448, opened 2022-08-08
Feat: add api register new vault collateral

bvotteler:
Add api to allow the connected account to register another collateral currency. Assuming the account id is already running a vault with another collateral currency.

-----
Signature in `VaultsAPI`:
```
    /**
     * Registers a new vault for the current account ID with a new collateral amount.
     * Only applicable if the connected account ID already has a running vault with a different collateral currency.
     *
     * Rejects with an Error if unable to register.
     *
     * @param collateralAmount The collateral amount to register the vault with - in the new collateral currency
     */
    registerNewCollateralVault(collateralAmount: MonetaryAmount<CollateralCurrencyExt>): Promise<void>;
```

### DIFF
--- a/src/parachain/vaults.ts
+++ b/src/parachain/vaults.ts
@@ -37,6 +37,8 @@ import {
 import { RewardsAPI } from "./rewards";
 import { BalanceWrapper, UnsignedFixedPoint } from "../interfaces";
 import { AssetRegistryAPI, SystemAPI } from "./index";
+import { ApiTypes, AugmentedEvent, SubmittableExtrinsic } from "@polkadot/api/types";
+import { ISubmittableResult, AnyTuple } from "@polkadot/types/types";
 
 /**
  * @category BTC Bridge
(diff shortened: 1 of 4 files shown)


src/parachain/vaults.ts
```diff
@@ -346,6 +357,32 @@ export class DefaultVaultsAPI implements VaultsAPI {
         return this.wrappedCurrency;
     }
 
+    async registerNewCollateralVault(collateralAmount: MonetaryAmount<CollateralCurrencyExt>): Promise<void> {
+        // check the vault account is set
+        const vaultAccount = this.transactionAPI.getAccount();
+        if (vaultAccount === undefined) {
+            return Promise.reject("Vault account must be set in the vaults API");
```
savudani8, line 364: Technically the account is set globally in the `interbtcapi` instance (`transactionAPI` is shared by reference in all the APIs), so this error message could be misleading

savudani8 (approved):
LGTM!

src/parachain/vaults.ts
bvotteler: > Technically the account is set globally in the interbtcapi instance (transactionAPI is shared by reference in all the APIs), so this error message could be misleading

Good catch.
I confused myself there and ended up rewriting the message incorrectly. 
I would like to indicate that the error originated inside vaults API where it couldn't be retrieved.

I'll add that to my refactoring work (a collection of non-blocking PR feedback items)

PR #452, opened 2022-08-10
Refactor: PR feedback items

bvotteler:
A collection of small changes suggested during pull request reviews.

### DIFF
--- a/test/integration/parachain/staging/sequential/replace.test.ts
+++ b/test/integration/parachain/staging/sequential/replace.test.ts
@@ -29,7 +29,7 @@ import { assert, expect } from "../../../../chai";
 import { issueSingle } from "../../../../../src/utils/issueRedeem";
 import { currencyIdToMonetaryCurrency, newAccountId, newVaultId, WrappedCurrency } from "../../../../../src";
 import { MonetaryAmount } from "@interlay/monetary-js";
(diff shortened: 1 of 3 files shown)


test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -86,27 +86,29 @@ describe("replace", () => {
...
+                    true,
+                    APPROX_FIFTEEN_BLOCKS_MS
+                );
+                await callWith(interBtcAPI, vault_3, async () =>
+                    interBtcAPI.replace.request(replaceAmount, collateralCurrency)
+                );
 
-                assert.isTrue(
-                    foundEvent,
+                await expect(foundEventPromise).to.eventually.be.equal(
```
savudani8, line 110: This is nice, I wonder though if the change doesn't reintroduce the race condition you tried to avoid when previously you batched the Promises in `Promise.all`. You can still do
```TypeScript
await Promise.all([
    callWith(interBtcAPI, vault_3, async () =>
        interBtcAPI.replace.request(replaceAmount, collateralCurrency)
    ),
    expect(foundEventPromise).to.eventually.be.equal( ... )
]);
```

savudani8 (approved):

PR #457, opened 2022-08-17
fix (test): should getFees test.

bvotteler:
Modify and reenable "should getFees" test in `vaults.test.ts`.
Add missing initializations of vault annuity account to make the test work again.


theodorebugnet (approved):

theodorebugnet:
LGTM - this might actually be a candidate test for being replaced with an RPC, but until then looks good

PR #458, opened 2022-08-17
feat: add electrs helper to get tx by largest payment

gregdhill:

There may be multiple Bitcoin transactions corresponding to an issue payment, if the user needs to "manually execute" this we should always show the highest amount not the oldest payment since a Vault may trick a user into executing the smaller amount.

We can probably remove or deprecate `getOldestPaymentToRecipientAddressTxId` later.

Not sure if we need to adjust `getTxIdByOpReturn` - would depend on where that is used.

### DIFF
--- a/src/external/electrs.ts
+++ b/src/external/electrs.ts
@@ -98,7 +98,19 @@ export interface ElectrsAPI {
      */
     getTxIdByOpReturn(opReturn: string, recipientAddress?: string, amount?: BitcoinAmount): Promise<string>;
     /**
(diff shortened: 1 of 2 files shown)


bvotteler (approved):
LGTM

bvotteler:
Just a few questions to help clarify wrt the TODO comments (paging):
Do we need a task somewhere for that? 
I'm not sure where that change needs to be made. After a quick following of the trail, it seems like a combination of OpenApi definitions in esplora and the implementations... somewhere else?

PR #464, opened 2022-08-24
feat: add method to get exchange rate for liquidation

bvotteler:
New method for `VaultsAPI`:
```
    /**
     * Get the target exchange rate at which a vault will be forced to liquidate, given its
     * current locked collateral and issued as well as to be issued tokens.
     *
     * @param vaultAccountId The vault's account ID
     * @param collateralCurrency The collateral currency for the vault with the account id above
     * @returns The theoretical collateral per wrapped currency rate below which the vault would be liquidated.
     *  Returns undefined if a value cannot be calculated, eg. if the vault has no issued tokens.
     */
    getExchangeRateForLiquidation(
        vaultAccountId: AccountId,
        collateralCurrency: CollateralCurrencyExt
    ): Promise<Big | undefined>;
```

### DIFF
--- a/src/parachain/vaults.ts
+++ b/src/parachain/vaults.ts
@@ -34,6 +34,7 @@ import {
     CollateralCurrencyExt,
     WrappedCurrency,
     GovernanceCurrency,
+    CurrencyExt,
 } from "../types";
 import { RewardsAPI } from "./rewards";
 import { UnsignedFixedPoint } from "../interfaces";
@@ -527,12 +528,7 @@ export class DefaultVaultsAPI implements VaultsAPI {
                 this.getTotalIssuedAmount(),
                 this.getIssuedAmount(vaultAccountId, collateralCurrency),
                 this.backingCollateralProportion(vaultAccountId, nominatorId, collateralCurrency),
-                (
-                    await this.tokensAPI.balance(
-                        await currencyIdToMonetaryCurrency(this.assetRegistryAPI, vault.id.currencies.collateral),
-                        vaultAccountId
-                    )
-                ).reserved,
+                this.getLockedCollateral(vault),
                 this.api.consts.timestamp.minimumPeriod,
             ]);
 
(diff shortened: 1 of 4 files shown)


src/parachain/vaults.ts
```diff
@@ -950,4 +959,31 @@ export class DefaultVaultsAPI implements VaultsAPI {
...
+        const issuedTokens = vaultExt.getBackedTokens();
+
+        if (issuedTokens.toBig().eq(0)) {
+            return undefined;
+        }
+
+        // calculate the theoretical exchange rate at which the vault would be (close to) liquidated
+        const liquidationRateCollaterallPerWrapped = lockedCollateral
+            .toBig()
+            .div(issuedTokens.toBig().mul(liquidationRateThreshold));
```
gregdhill, line 986: So the calculation we do on the parachain is `collateral / exchange_rate / threshold = wrapped`.

Plugging in an example to remind myself:
- exchange rate is 2333 DOT per BTC (converted to planck / sat this is `2333 * (10**10 / 10**8)`)
- if we lock 2333 DOT we can mint `(2333 * 10**10) / 233300 / 2.6 = 0.38 BTC`
- or put another way, to mint 1 BTC we need `(1 * 10**8) * 233300 * 2.6 = 6065.8 DOT`
- so to get the exchange rate for liquidation we can do `(6065.8 * 10**10) / (1 * 10**8 * 2.6) = 233300`
- or for liquidation `(6065.8 * 10**10) / (1 * 10**8 * 1.1) = ~551436.36`
- so the liquidation exchange rate in this case is 5514.36 DOT per BTC (`551436.36 / (10**10 / 10**8)`)

Therefore the calculation we need to do is `exchange_rate = collateral / (wrapped * threshold)`.

gregdhill: tl;dr this implementation looks fine

gregdhill:
```suggestion
        const liquidationRateCollateralPerWrapped = lockedCollateral
            .toBig()
            .div(issuedTokens.toBig().mul(liquidationRateThreshold));
```

gregdhill (approved):
LGTM, typo in variable name but approved anyway

PR #465, opened 2022-08-25
feat: add script for generating x-tokens treasury preimage

gregdhill:

Simple script for generating the preimage for xTokens transfer proposals.


scripts/pool-setup.ts
```diff
@@ -0,0 +1,161 @@
...
+        "https://polkadot.js.org/apps/?rpc=" +
+        encodeURIComponent(endpoint) +
+        "#/extrinsics/decode/" +
+        extrinsic.method.toHex();
+    console.log("Url:", url);
+    console.log("");
+}
+
+function palletIdToAccountId(api: ApiPromise, palletId: any): string {
+    const accountId = Buffer.concat([
```
bvotteler, line 62: Nit: This could be condensed to:
```
Buffer.concat([
    Buffer.from("modl"),
    Buffer.from(palletId)
  ],
  32
);
```
ie. we can skip the calculation with magic numbers.

bvotteler: Example test case:
```
    it.only("buffer concat", () => {
        const abc = Buffer.from("abc");
        const one23 = Buffer.from("123");
        const buf = Buffer.concat([abc, one23], 32);

        // kept length
        expect(buf.length).to.be.eq(32);
        // filled overallocated bytes with zeroes
        expect(buf.readBigUInt64BE(abc.length + one23.length)).to.be.eq(BigInt(0));
    });
```
^ passes for me

bvotteler (approved):
LGTM

scripts/pool-setup.ts
gregdhill: Thanks! I wonder if we should actually move this code into the utils and actually add those unit tests?

bvotteler: > Thanks! I wonder if we should actually move this code into the utils and actually add those unit tests?

We can move the code to the utils if there is a shared use case now, or in future for that matter. 
I'm not aware if any of the other scripts do something similar.

PR #466, opened 2022-09-01
chore: replace several in-place implementations with RPC calls to simplify the lib, in escrow and rewards modules

theodorebugnet:
Draft until parachain is runtime-upgraded with the new RPCs


peterslany (approved):
Looks good to me :+1:

PR #469, opened 2022-09-06
Feat: add getter for foreign asset collateral currencies

bvotteler:
New method for `AssetRegistryApi`:
```
    /**
     * Get all foreign assets which have a registered collateral ceiling, meaning they can be used as collateral currency.
     *
     * @returns All foreign assets that have been registered as collateral currency
     */
    getCollateralForeignAssets(): Promise<Array<ForeignAsset>>;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.16.7",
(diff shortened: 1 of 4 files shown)


peterslany (approved):
Good job adding this so quickly :) Looks good to me :+1:

PR #470, opened 2022-09-12
feat: provide methods to build extrinsics

bvotteler:
Provide APIs to get raw submittable extrinsics (without submitting) in order to allow for easier batching.

This allows consumers of the lib to better construct and submit a collection of extrinsics/transactions.

-----
Example:
A vault could disable issuing, and add more collateral in one batch.
```
// disable issuing for vault with DOT collateral
const disableIssuing = vaultsApi.buildAcceptNewIssuesExtrinsic(Polkadot, false);
// we want to add 42 DOT
const collateralAmount = newMonetaryAmount(42, Polkadot, true);
// create extrinsic to add collateral
const addCollateral = vaultsApi.buildDepositCollateralExtrinsic(collateralAmount);
// create batch extrinsic (treat extrinsics atomically - all or none succeed)
const batchExtrinsic = transactionAPI.buildBatchExtrinsic([disableIssuing, addCollateral], true);
// submit batch, and wait for Promise to resolve/reject
await transactionAPI.sendLogged(batchExtrinsic);
```


savudani8 (commented):
I think that rather than having lib users call `buildBatchExtrinsic` directly, the library should provide a `BatchTransactionBuilder` class that has methods such as `withIssueRequest(...)`, calling the builder method and storing it in the class. Then a `send()` method would build the batch extrinsic and submit it. (I don't think it would make sense to batch multiple `batch` txs so `withBatch` isn't required in the builder interface). Such a builder class would also need to store success events and ensure they occur in order. Maybe something for the next PR though?

bvotteler:
> Such a builder class would also need to store success events and ensure they occur in order. Maybe something for the next PR though?

Definitely something to think about for future work.
There are a couple of things to consider if a builder is the best approach - even though I really like the way the syntax would look like.
Such as if we can simply construct it with an existing InterBtcAPI instance wherever we need it, or if we want to allow passing in various other APIs at the cost of runtime errors.

There might also be other options to ensure success events are in order. Although I'm not sure yet how the lib could enforce a different order if the parachain does not execute the batched extrinsic in order of the provided array/vector.

savudani8 (approved):

PR #472, opened 2022-09-16
Feat: fetch collateral currencies from chain

bvotteler:
Fetch collateral currencies based on parachain data, rather than the previously hard-coded token currencies.
The method will now also return foreign assets, as opposed to hard-coded tokens (KINT, INTR, KSM, DOT, etc) only.

------

In `src/utils/currency.ts`
Marked method as deprecated (note of planned removal in the next minor version bump (1.17.0): 
`getCorrespondingCollateralCurrencies(governanceCurrency: GovernanceCurrency): Array<CollateralCurrencyExt>`
To be used instead:
`getCollateralCurrencies(api: ApiPromise, assetRegistry: AssetRegistryAPI): Promise<Array<CollateralCurrencyExt>>`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.16.9",
+  "version": "1.16.10",
   "description": "JavaScript library to interact with interBTC and kBTC",
   "main": "build/src/index.js",
   "typings": "build/src/index.d.ts",
--- a/src/utils/currency.ts
+++ b/src/utils/currency.ts
@@ -96,6 +96,25 @@ export function toVoting(governanceCurrency: GovernanceCurrency): Currency {
 }
 
 /**
+ * Get the collateral currencies (excluding foreign assets) associated with the governance currency.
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.16.9",
```
nud3l, line 3: If it's a breaking change, might be good to bump the minor version.

bvotteler: > If it's a breaking change, might be good to bump the minor version.

You got me. :)

I was trying to "save the bump" for the upcoming removal of some APIs.
But yes, I agree. 
I'm thinking I might deprecate the old API (rather than remove it), with a note that it will be gone in the next minor version. The UI team is expecting this change, and hopefully that isn't too short notice for other consumers we're not aware of.

gregdhill (approved):
Lgtm (apart from the version bump @nud3l suggested)

package.json
bvotteler: Marking as resolved after adding the old method back, marked as deprecated. Therefore, no breaking changes (yet)

PR #477, opened 2022-10-10
Peter/feat interlend module methods

peterslany:
- Adding base methods and types to enable retrieving user's lend and borrow positions for Interlend. These data are mocked for now. The response is an object that maps asset ticker to `LoanPosition` object. Collateralization (health of the loan positions) will be computed on the UI to prevent unnecessary calls to parachain - we will be fetching USD price in the UI, so we can use it to compute the collateralization of loans as well.

- Adding method to retrieve all assets enabled for loans with additional information about apy, liquidity, loan capacity, etc. The returned data is object that maps currency ticker to `LoanAsset` type object.


src/types/loans.ts
```diff
@@ -0,0 +1,38 @@
+import { MonetaryAmount } from "@interlay/monetary-js";
+import Big from "big.js";
+import { CurrencyExt } from "./currency";
+
+interface LoanReward{
+    currency: CurrencyExt;
+    apy: Big;
```
bvotteler, line 7: Might be a good idea to indicate in a comment if the `apy` value is expected in percent or as a decimal rate

src/types/loans.ts
```diff
@@ -0,0 +1,38 @@
...
+
+interface LoanPosition {
+    currency: CurrencyExt;
+    amount: MonetaryAmount<CurrencyExt>;
+    earnedReward: MonetaryAmount<CurrencyExt> | null; // null if rewards are not enabled.
+}
+
+interface LendPosition extends LoanPosition {
+    isCollateral: boolean;
+    earnedInterest: MonetaryAmount<CurrencyExt>;
```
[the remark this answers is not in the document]
bvotteler: Actually, ignore that. I confused LoanPosition with LendPosition.

src/types/loans.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+type BorrowPosition = LoanPosition;
+
+interface LoanAsset {
+    currency: CurrencyExt;
+    lendApy: Big;
+    borrowApy: Big;
```
danielsimao, line 26: There could be an asset that cannot be lend but could be borrowed. This Scenario happens for USDT on Compound. Same thing could happen the other way around. Can be lended, but not borrowed. For example LP's.

danielsimao (changes requested):

src/types/loans.ts
peterslany: @danielsimao Actually in our case, that is not possible. Check this thread where we discussed it: https://discord.com/channels/745259537707040778/1026832208612306974/1026836371081539705

src/types/loans.ts
peterslany: Good idea, comment added - value will be in percentage.

bvotteler (approved):
lgtm

danielsimao (approved):

PR #484, opened 2022-10-20
fix: redeem expiry status

bvotteler:
Fix a bug where the utility method `parseRedeemRequestStatus` returns a pending status when it should return expired and vice versa.

### DIFF
--- a/src/utils/encoding.ts
+++ b/src/utils/encoding.ts
@@ -283,7 +283,7 @@ export function parseRedeemRequestStatus(
     // now check if pending is actually expired
     const maxPeriod = Math.max(req.period.toNumber(), redeemPeriod);
     const openTime = req.opentime.toNumber();
-    if (openTime + maxPeriod > activeBlockCount) {
+    if (openTime + maxPeriod <= activeBlockCount) {
         return RedeemStatus.Expired;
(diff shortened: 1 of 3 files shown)


src/utils/encoding.ts
```diff
@@ -283,7 +283,7 @@ export function parseRedeemRequestStatus(
     // now check if pending is actually expired
     const maxPeriod = Math.max(req.period.toNumber(), redeemPeriod);
     const openTime = req.opentime.toNumber();
-    if (openTime + maxPeriod > activeBlockCount) {
+    if (openTime + maxPeriod <= activeBlockCount) {
```
peterslany, line 286: Shouldn't this be unequal comparison - only `<` not `<=`? Because I think that in case when current block is the same as the expiration block, the request is still valid and becomes expired only in the block after:
https://github.com/interlay/interbtc/blob/c88fc96634160ed585fb8467a98992f52b622a69/crates/security/src/lib.rs#L

Please correct me if I'm wrong.

bvotteler: Yes, you are right. Good point. Thanks!

peterslany (approved):

PR #494, opened 2022-11-15
Peter/interlend  my borrow positions impl

peterslany:
Opening smaller PR for getter of my borrow positions, last data fetching method needed for the UI.

### DIFF
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -1,5 +1,5 @@
 import { AccountId } from "@polkadot/types/interfaces";
-import { KBtc, Kintsugi, MonetaryAmount } from "@interlay/monetary-js";
+import { Kintsugi, MonetaryAmount } from "@interlay/monetary-js";
 import {
     BorrowPosition,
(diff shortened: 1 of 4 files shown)


src/parachain/loans.ts
```diff
@@ -304,7 +313,45 @@ export class DefaultLoansAPI implements LoansAPI {
         };
     }
 
-    async getLendPositionsOfAccount(accountId: AccountId): Promise<Array<LendPosition>> {
+    _calculateEarnedDebt(borrowedAmount: Big, snapshotBorrowIndex: Big, currentBorrowIndex: Big): Big {
```
savudani8, line 316: Since debt is a liability rather than an asset, "earned" isn't the best wording here. I suggest renaming this to `_calculateAccumulatedDebt`.

savudani8 (approved):

src/parachain/loans.ts
peterslany: good suggestion, thanks!

PR #497, opened 2022-11-15
Peter/interlend borrow repay impl

peterslany:
Adding implementation and simple tests for borrow, repay and claimReward extrinsics.

### DIFF
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -130,48 +130,40 @@ export interface LoansAPI {
     disableAsCollateral(underlyingCurrency: CurrencyExt): Promise<void>;
 
     /**
-     * Claim subsidy reward for specified currency.
-     *
-     * @param currency Currency for which to claim reward.
-     * @throws If no position exists for `currency` and account.
-     */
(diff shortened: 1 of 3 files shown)


src/parachain/loans.ts
```diff
@@ -335,6 +327,10 @@ export class DefaultLoansAPI implements LoansAPI {
         const currentBorrowIndex = Big(decodeFixedPointType(marketStatus[6]));
         const accumulatedDebt = this._calculateAccumulatedDebt(borrowedAmount, snapshotBorrowIndex, currentBorrowIndex);
 
+        if (borrowedAmount.eq(0)) {
+            return null;
+        }
```
bvotteler, line 323: Suggestion: Move this up a bit to avoid calculating numbers that don't matter if we return `null`

bvotteler (approved):
Looks fine to me. 

I do suggest moving the early `null` return up a bit (see other comment), but that isn't blocking. We can also get back to this later.

src/parachain/loans.ts
peterslany: Thanks, that's very good catch, actually it also resolves bug caused by zero division that I found :raised_hands:

PR #500, opened 2022-11-18
refactor(interlend): add separate method for getting acrruedReward

peterslany:
Refactoring rewards by putting them outside of the `LoanPosition` type and adding separate method to obtain them.


src/parachain/loans.ts
```diff
@@ -462,13 +461,19 @@ export class DefaultLoansAPI implements LoansAPI {
         return amountPerBlock.mul(blocksPerYear);
     }
 
-    _getSubsidyReward(amount: Big): MonetaryAmount<CurrencyExt> | null {
+    async _getRewardCurrency(): Promise<CurrencyExt> {
+        const rewardCurrencyId = this.api.consts.loans.rewardAssetId;
```
bvotteler, line 465: nice

test/integration/parachain/staging/sequential/loans.test.ts
```diff
@@ -373,6 +371,12 @@ describe("Loans", () => {
         });
     });
 
+    describe("getAccruedRewardsOfAccount", () => {
+        it("should return correct amount of reward", () => {
```
bvotteler, line 375: We should check if this test is needed as integration test.
We might get away with just testing our calcs as unit tests instead.

bvotteler (approved):
LGTM, good idea to extract that.

src/parachain/loans.ts
```diff
@@ -462,13 +461,19 @@ export class DefaultLoansAPI implements LoansAPI {
...
+
+        return currencyIdToMonetaryCurrency(this.assetRegistryAPI, this, rewardCurrencyId);
+    }
+
+    _getSubsidyReward(amount: Big, rewardCurrency: CurrencyExt): MonetaryAmount<CurrencyExt> | null {
```
peterslany, line 470: Realized we don't need to pass the governance currency into the constructor as the runtime contains the reward currency already. :)

src/parachain/loans.ts
```diff
@@ -451,8 +450,8 @@ export class DefaultLoansAPI implements LoansAPI {
 
         // Return rate per 1 UNIT of underlying currency and compute APR
         // on UI where all exchange rates are available.
-        const lendRewardPerUnit = lendRewardPerPool.div(totalLiquidity);
-        const borrowRewardPerUnit = borrowRewardPerPool.div(totalBorrows);
+        const lendRewardPerUnit = totalLiquidity.eq(0) ? lendRewardPerPool : lendRewardPerPool.div(totalLiquidity);
+        const borrowRewardPerUnit = totalBorrows.eq(0) ? borrowRewardPerPool : borrowRewardPerPool.div(totalBorrows);
```
peterslany, line 454: Spotted and fixed the bug of zero division in some edge cases here.

test/integration/parachain/staging/sequential/loans.test.ts
peterslany: Hmm true, it just returns value from the parachain so no calculations to test on that front :+1:

PR #501, opened 2022-11-24
feat(loans): add liquidateBorrow extrinsic

peterslany:
Adding a liquidate borrow position extrinsic to the loans module.

### DIFF
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -164,6 +164,21 @@ export interface LoansAPI {
      * @param underlyingCurrency Currency to repay.
      */
     repayAll(underlyingCurrency: CurrencyExt): Promise<void>;
+
+    /**
+     * Liquidates borrow position for exchange of collateral.
+     *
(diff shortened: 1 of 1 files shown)


src/parachain/loans.ts
```diff
@@ -164,6 +164,21 @@ export interface LoansAPI {
...
+     * @param borrower AccountId of borrower whose position will be liquidated.
+     * @param liquidationCurrency Currency of position that will be liquidated.
+     * @param repayAmount Amount to be repaid.
+     * @param collateralCurrency Collateral currency which will be claimed by liquidator.
+     */
+    liquidateBorrowPosition(
+        borrower: AccountId,
+        liquidationCurrency: CurrencyExt,
+        repayAmount: MonetaryAmount<CurrencyExt>,
+        collateralCurrency: CurrencyExt
```
peterslany, line 180: Using `MonetaryAmount` here to stay consistent with other methods of the module, @savudani8 let me know if this should be an issue for the grant program as it create additional dependency on `monetary-js`

PR #502, opened 2022-11-25
Feat: Loans module

peterslany:
Merging loans branch back into master with updated runtime and client versions.


src/parachain/loans.ts
```diff
@@ -0,0 +1,625 @@
...
+     * @returns Array of borrow positions of account.
+     */
+    getBorrowPositionsOfAccount(accountId: AccountId): Promise<Array<BorrowPosition>>;
+
+    /**
+     * Get all loan assets.
+     *
+     * @returns Array of all assets that can be lent and borrowed.
+     * @remarks Method could be refactored to compute APR in lib if we can get underlyingCurrency/rewardCurrency exchange rate,
+     * but is it safe to assume that exchange rate for btc/underlyingCurrency will be
```
savudani8, line 46: it is safe to assume that, yes

src/parachain/loans.ts
```diff
@@ -0,0 +1,625 @@
...
+    /**
+     * Same as `withdraw`, but exits full position.
+     *
+     * @param underlyingCurrency Currency to fully withdraw.
+     */
+    withdrawAll(underlyingCurrency: CurrencyExt): Promise<void>;
+
+    /**
+     * Enable lend position of account as collateral for borrowing.
+     *        await this._checkMarketState(underlyingCurrency, "withdraw");
```
savudani8, line 107: leftover code here?

src/parachain/loans.ts
```diff
@@ -0,0 +1,625 @@
...
+            this.api.query.loans.accountDeposits(lendTokenId, accountId),
+            this.api.rpc.loans.getMarketStatus(underlyingCurrencyId),
+        ]);
+
+        const startingExchangeRate = decodeFixedPointType(accountEarned.exchangeRatePrior);
+        const currentExchangeRate = decodeFixedPointType(currentMarketStatus[2]);
+        const earnedPrior = decodeFixedPointType(accountEarned.totalEarnedPrior);
+        const earnedInterest = currentExchangeRate
+            .sub(startingExchangeRate)
+            .mul(underlyingCurrencyAmount) // TODO: check whether to use underlying or lendtoken balance here
```
savudani8, line 271: This should be the lendtoken

savudani8 (approved):

src/parachain/loans.ts
peterslany: Yes, probably pasted there by mistake.

src/parachain/loans.ts
peterslany: Thanks! I updated the computation :)

src/parachain/loans.ts
peterslany: Good to know! This can be refactored later if there is a need - for now I think the current solution works well enough.

PR #525, opened 2023-01-09
WIP Peter/amm feat pools and swap

peterslany:
Adding 
- Standard and Stable pool, Trade classes that help with routing
- `getLiquidityPools` method to get all pools
- `getOptimalTrade` method that returns best trade path for provided available pools, input amount and output currency


src/parachain/amm/liquidity-pool/stable.ts
```diff
@@ -0,0 +1,280 @@
+import { CurrencyExt, StableLPToken } from "../../../types";
```
peterslany, line 1: Inspired by zenlink-dex package, logic is translated from their `sdk-core`

src/parachain/amm/liquidity-pool/stable.ts
```diff
@@ -0,0 +1,280 @@
...
+        const D0 = this._getD(this.xp, amp);
+
+        const newBalances = this.pooledCurrencies.map((balance, i) =>
+            deposit ? balance.add(sortedAmounts[i]) : balance.sub(sortedAmounts[i])
+        );
+        const D1 = this._getD(this._xp(newBalances), amp);
+
+        if (this.totalSupply.isZero()) {
+            // TODO: check is D1 in base or atomic denomination??
+            // if not in base divide by decimals first
```
bvotteler, line 214: If it is in atomic, you can use the lib utility function `newMonetaryAmount()` which accepts atomic (or base)

src/parachain/amm/liquidity-pool/stable.ts
bvotteler: Good job so far :)
And a big thank you.

src/parachain/amm/trade/trade.ts
```diff
@@ -0,0 +1,59 @@
...
+        return this.outputAmount.gte(anotherTrade.outputAmount);
+    }
+
+    /**
+     * Get minimum output amount for trade with provided maximum allowed slippage.
+     *
+     * @param {number} maxSlippage Maximum slippage in percentage.
+     * @returns {MonetaryAmount<CurrencyExt>} Minimum output amount of trade allowed with provided slippage.
+     */
+    public getMinimumOutputAmount(maxSlippage: number): MonetaryAmount<CurrencyExt> {
```
bvotteler, line 50: Does this need validation for sensible slippage numbers?

e.g. We can get a negative minimum amount if `maxSlippage > 100`

src/types/currency.ts
```diff
@@ -36,10 +36,14 @@ export type CollateralIdLiteral =
 const CollateralCurrency = [Polkadot, Kusama, Interlay, Kintsugi] as const;
 type CollateralCurrency = typeof CollateralCurrency[number];
 
-export type ForeignAsset = Currency & {foreignAsset: { id: number; coingeckoId: string }};
-export type LendToken = Currency & {lendToken: { id: number }}; 
+export type ForeignAsset = Currency & { foreignAsset: { id: number; coingeckoId: string } };
+export type LendToken = Currency & { lendToken: { id: number } };
+export type StandardLPToken = Currency & { lpToken: { token0: CurrencyExt; token1: CurrencyExt } };
```
bvotteler, line 41: I think if we want to avoid the circular reference here.
`StandardLpToken` can contain up to 2 `CurrencyExt` which can be an `LpToken` which can be a `StandardLPToken`

I was under the impression `tokenX` can only be a native token, foreign asset, or stable pool id.
See type definition in [interbtc-types definitions](https://github.com/interlay/interbtc-types/blob/3d4736fb3a1ca6b51ace54f8995134d3d98d9690/definitions.json#L12-L27)

bvotteler, line 41: Up for discussion: I would prefer `<something optional>LpToken` with lower case `p` rather than capitalized `P`. Even though it stands for `LiquidityPool` (I think?) but makes it just more readable in my opinion.

src/utils/currency.ts
```diff
@@ -165,6 +176,7 @@ export function getCurrencyIdentifier(currency: CurrencyExt): CurrencyIdentifier
     return { token: currency.ticker };
 }
 
+// TODO: add StandardLPToken and StableLPToken
```
bvotteler, line 179: Something to consider for the TODOs:
Would it make sense to capture those in one or several issues? Just so we don't lose track with the many small changes.

src/parachain/amm.ts
```diff
@@ -0,0 +1,443 @@
...
+        return [...standardPools, ...stablePools];
+    }
+
+    async swap(
+        trade: Trade,
+        minimumAmountOut: MonetaryAmount<CurrencyExt>,
+        recipient: AddressOrPair,
+        deadline: number | string
+    ): Promise<void> {
+        //TODO
```
bvotteler, line 429: Would throwing not implemented make sense for this method (and the next two) make sense?
If so, that would make it consistent with the other unimplemented ones.

src/parachain/amm/liquidity-pool/stable.ts
peterslany: Yes thanks for the reminder!

src/parachain/amm/liquidity-pool/stable.ts
peterslany: Thank you for the review!

src/types/currency.ts
peterslany: This is good suggestion, actually the runtime uses lowercase `p` as well, so I changed it :+1:

peterslany: This is good suggestion, actually the runtime uses lowercase p as well, so I changed it +1

peterslany: Thanks for noticing this :pray:

src/utils/currency.ts
peterslany: Yes, I already created several issues.

bvotteler (approved):

PR #527, opened 2023-01-11
chore: fix lib path in tsconfig

peterslany:
Fixing lib path in `tsconfig.json`.


test/unit/parachain/asset-registry.test.ts
```diff
@@ -134,7 +134,11 @@ describe("DefaultAssetRegistryAPI", () => {
 
     describe("getCollateralForeignAssets", () => {
         // only id matters for these tests
-        const mockForeignAssets = [<ForeignAsset>{ foreignAsset: {id: 1} }, <ForeignAsset>{ foreignAsset: {id: 2} }, <ForeignAsset>{ foreignAsset: {id: 3 }}];
+        const mockForeignAssets = [
```
bvotteler, line 137: Random thought: We should probably get a consistent line width eslint/prettier config setup between the various typescript devs to avoid these types of changes.

bvotteler (approved):

test/unit/parachain/asset-registry.test.ts
peterslany: @bvotteler Good idea, actually I looked at eslint config and we have this rule:

https://github.com/interlay/interbtc-api/blob/9b7c0b6e3db6d64e394cd09ef45edabdcfa8fdc5/.eslintrc.json#L24-L29
Not sure why it was not enforced in the same way :thinking:

bvotteler: I think I removed that because it kept disagreeing with prettier.

PR #528, opened 2023-01-12
chore: script to check surplus balance of sovereign account

sander2:
The script is a bit janky but it works so I don't feel inclined to spend more time on it. This script checks the surplus balance of our chain's sovereign accounts on the relay chain, which we need to know when we want to open new xcm channels


bvotteler (commented):
LGTM

bvotteler (approved):

PR #537, opened 2023-01-18
fix: rename lib LpToken -> LpCurrency

peterslany:
Rename `LpToken` currency type union to `LpCurrency` to avoid having same name as runtime `LpToken` type.


bvotteler (approved):
Not ideal, but better than type name clashes.

LGTM

PR #538, opened 2023-01-18
Feat: use rewards pallet rpc methods for reward rate estimates

### LINKED ISSUE #530: Use new rewards rpc
We can remove the (now) duplicate logic in `DefaultEscrowAPI.getRewardEstimate` and `DefaultVaultAPI.getBlockRewardAPY`.

Parachain version 1.21.1 provides two new rpc apis in `rpc.rewards`:
- `.estimateEscrowRewardRate`, and
- `.estimateVaultRewardRate`

See PR here: https://github.com/interlay/interbtc/pull/842/files#diff-3e8f65b5a19c0ab3c6a342a1ed4999b53f61b95ce355b547a6d80b868a60b9bbR23-R28

bvotteler:
Integrating the new RPCs for calculating the APY

Rust implementation: https://github.com/interlay/interbtc/blob/042085bc8def193657e55bb98021da7f14b2276d/crates/reward/rpc/src/lib.rs#L48-L58

Replacing the implementations here:
- https://github.com/interlay/interbtc-api/blob/4768cdda68477e7ea1193bfa4f82ddef4d529bfc/src/parachain/escrow.ts#L205
- https://github.com/interlay/interbtc-api/blob/4768cdda68477e7ea1193bfa4f82ddef4d529bfc/src/parachain/vaults.ts#L628

⚠️ Note: Includes a required a bump to parachain version `1.21.3` and clients version `1.19.1`

Resolves #530


src/parachain/escrow.ts
```diff
@@ -210,29 +210,38 @@ export class DefaultEscrowAPI implements EscrowAPI {
...
+        let nullableBlockToEnd: bigint | null = null;
+        if (blockLockTimeExtension) {
+            const currentEndBlock = !stakedBalance.amount.isZero() ? BigInt(stakedBalance.endBlock) : BigInt(0);
+            nullableBlockToEnd = currentEndBlock + BigInt(blockLockTimeExtension);
+        }
+
+        const rawRewardRate = await this.api.rpc.reward.estimateEscrowRewardRate(
+            accountId,
+            nullableAmountToCheck,
+            nullableBlockToEnd
```
gregdhill, line 233: Okay so probably I should've been more clear about this before (I'll add a comment to the RPC later) but actually these are the amount & length to extend. It already considers the staked balance and end lock time, so you can make the above function even simpler!

src/parachain/escrow.ts
```diff
@@ -210,29 +210,38 @@ export class DefaultEscrowAPI implements EscrowAPI {
...
         );
+
+        // annual reward rate
+        const rewardRate = decodeFixedPointType(rawRewardRate);
+        // reward amount
+        const rewardAmount = newMonetaryAmount(atomicStakedAmount, this.governanceCurrency).mul(rewardRate);
```
gregdhill, line 239: Ah I didn't realize you also needed the concrete reward amount, so this is actually wrong because in the parachain `rate = received / total_locked` where `total_locked` is `stakedBalance + amount`.

gregdhill: So to get the reward amount it would be `received = rate * (stakedBalance + amount)`

gregdhill (changes requested):

src/parachain/escrow.ts
bvotteler: > Okay so probably I should've been more clear about this before (I'll add a comment to the RPC later) but actually these are the amount & length to extend. It already considers the staked balance and end lock time, so you can make the above function even simpler!

That makes me unreasonable happy. On it.

src/parachain/escrow.ts
bvotteler: > So to get the reward amount it would be received = rate * (stakedBalance + amount)

I think I got it covered here?
https://github.com/interlay/interbtc-api/pull/538/files/592f0346f2cf87b19a23377460e019580f056cc0#diff-4573174c08a415f939d2d03353df39634595439f0f5a301e7c4f2ee7ab58d55cR219

gregdhill (approved):

src/parachain/escrow.ts
gregdhill: Ah right I missed that

PR #539, opened 2023-01-18
feat(amm): add swap extrinsic

### LINKED ISSUE #533: AMM: Implement swap extrinsic
Implement AMM module `swap` method for making trades.

peterslany:
Resolves #533

### DIFF
--- a/src/parachain/amm.ts
+++ b/src/parachain/amm.ts
@@ -464,7 +464,7 @@ export class DefaultAMMAPI implements AMMAPI {
             deadline
         );
 
-        await this.transactionAPI.sendLogged(swapExtrinsic, this.api.events.zenlinkStableAmm.CurrencyExchange);
+        await this.transactionAPI.sendLogged(swapExtrinsic, this.api.events.zenlinkStableAmm.CurrencyExchange, true);
     }
 
     async swap(
--- a/src/parachain/amm.ts
+++ b/src/parachain/amm.ts
@@ -27,6 +27,9 @@ import {
     findBestTradeRecursively,
     StandardLiquidityPool,
     StableLiquidityPool,
+    isStableMultiPathElement,
+    encodeSwapParamsForStandardPoolsOnly,
+    encodeSwapParamsForStandardAndStablePools,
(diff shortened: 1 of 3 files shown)


src/parachain/amm.ts
```diff
@@ -420,14 +423,62 @@ export class DefaultAMMAPI implements AMMAPI {
...
+        );
+
+        await this.transactionAPI.sendLogged(swapExtrinsic, this.api.events.zenlinkStableAmm.CurrencyExchange, true);
+    }
+
     async swap(
         trade: Trade,
         minimumAmountOut: MonetaryAmount<CurrencyExt>,
-        recipient: AddressOrPair,
+        recipient: string,
```
bvotteler, line 473: Does the recipient string represent an Account?
If so, I would prefer if it was typed more strongly.

src/parachain/amm/encoding.ts
```diff
@@ -0,0 +1,76 @@
...
+    )[];
+} {
+    const amountIn = trade.inputAmount._rawAmount.toString();
+    const amountOutMin = minimumAmountOut._rawAmount.toString();
+    const path = trade.path.map((pathElement) =>
+        isStableMultiPathElement(pathElement)
+            ? {
+                  Stable: {
+                      poolId: pathElement.pool.poolId,
+                      basePoolId: isStableMetaMultiPathElement(pathElement) && pathElement.basePool.poolId,
```
bvotteler, line 58: What happens if `isStableMetaMultiPathElement()` returns false? 
Is any falsy value valid for `basePoolId`?

src/parachain/amm.ts
peterslany: Yes, it represent an account, updated :+1:

src/parachain/amm/encoding.ts
peterslany: I think this was mistake, good catch! Zenlink sdk returns undefined here, but now I changed it to explicit 0.

bvotteler (approved):
LGTM

PR #541, opened 2023-01-19
eat(amm): implement or mock all methods for getting pool data

peterslany:
Adding implementation for submethods that `getPools` depends on.


src/parachain/amm.ts
```diff
@@ -262,27 +263,27 @@ export class DefaultAMMAPI implements AMMAPI {
...
-        throw new Error("Method not implemented.");
-    }
+    ): Promise<[MonetaryAmount<CurrencyExt>, MonetaryAmount<CurrencyExt>]> {
+        const [token0Balance, token1Balance] = await Promise.all([
+            this.tokensAPI.balance(token0, pairAccount),
+            this.tokensAPI.balance(token1, pairAccount),
+        ]);
+        // TODO: check if transferable or free balance should be used here
+        const token0MonetaryAmount = token0Balance.transferable;
+        const token1MonetaryAmount = token1Balance.transferable;
```
peterslany, line 277: @gregdhill Is it correct to use only transferable balance here for the balance of the uniswap pool or should it be something else?

gregdhill: It can be either I think (assuming "transferable" is `free - locked` since we never lock) but I would advise you to use the free balance

Issue #535, opened 2023-01-16
AMM: Implement methods to calculate liquidity deposit/withdrawal amounts


peterslany:
Implement methods for AMM module:
- `getExpectedLiquidityDepositAmounts` - calculate how much LP tokens will the account get after depositing liquidity (both standard and stable pools - for stable pools need to take slippage into account)
- `getExpectedLiquidityWithdrawalAmounts` - calculate how much underlying tokens will the account get after withdrawing liquidity from pool (both standard and stable pools - for stable pools need to take into account proportion of underlying assets that will be withdrawn)

PR #544, opened 2023-01-23
Peter/amm calculate liquidity amounts

peterslany:
Adding class that provides calculations needed when depositing/withdrawing liquidity to/from pools.

Resolves https://github.com/interlay/interbtc-api/issues/535


bvotteler (approved):
LGTM

PR #545, opened 2023-01-23
feat: add util to compute block number in future

peterslany:
Adding method for computing what will be the actual block number in X seconds.

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -11,3 +11,5 @@ export const IGNORED_ERROR_MESSAGES = ["issue.IssueCompleted"];
 export const ACCOUNT_NOT_SET_ERROR_MESSAGE = "cannot sign transaction without setting account";
 
 export const MS_PER_YEAR = Big(86400 * 365 * 1000);
+
+export const BLOCK_TIME_SECONDS = 12;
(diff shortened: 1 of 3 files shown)


src/utils/constants.ts
```diff
@@ -11,3 +11,5 @@ export const IGNORED_ERROR_MESSAGES = ["issue.IssueCompleted"];
 export const ACCOUNT_NOT_SET_ERROR_MESSAGE = "cannot sign transaction without setting account";
 
 export const MS_PER_YEAR = Big(86400 * 365 * 1000);
+
+export const BLOCK_TIME_SECONDS = 12;
```
bvotteler, line 15: This is fine because it seems unlikely to change.

But we can also read the duration from the parachain's value.
```
const blockTimeMs = (await this.api.call.auraApi.slotDuration()).toNumber();
```

test/integration/parachain/staging/system.test.ts
```diff
@@ -38,9 +38,16 @@ describe("systemAPI", () => {
...
         await sudo(interBtcAPI, () => interBtcAPI.system.setCode(code));
     });
+
+    it("should getFutureBlockNumber", async () => {
+        const [currentBlockNumber, futureBlockNumber] = await Promise.all([
+            interBtcAPI.system.getCurrentBlockNumber(),
+            interBtcAPI.system.getFutureBlockNumber(120),
+        ]);
+
+        assert.isTrue(currentBlockNumber + 10 === futureBlockNumber);
```
bvotteler, line 51: I wonder if we can run into some race condition / slippage there.

I'd suggest adding some wiggle room with a combined

```
assert.isAtLeast(futureBlockNumber, currentBlockNumber + 9);
assert.isAtMost(futureBlockNumber, currentBlockNumber + 11);
```

Otherwise, if we're certain it must be 10 add a meaningful error message in case it isn't.

bvotteler (changes requested):

bvotteler (approved):
LGTM

Issue #534, opened 2023-01-16
AMM: Implement methods to add/remove liquidity


peterslany:
Implement methods for AMM module:
- `addLiquidity` to add liquidity to pool (for both standard and stable pools)
- `removeLiquidity` to remove liquidity from pool (for both standard and stable pools)

Note: if there is farm for the LP token it should be automatically deposited on liquidity addition and withdrawn on liquidity withdrawal

PR #547, opened 2023-01-24
Peter/amm implement liquidity extrinsics methods

peterslany:
Resolves https://github.com/interlay/interbtc-api/issues/534

- Adding implementation of methods to add and remove liquidity.
- Adding `StableLiquidityMetaPool` class to represent meta pools.

### DIFF
--- a/src/parachain/amm.ts
+++ b/src/parachain/amm.ts
@@ -198,6 +198,9 @@ export class DefaultAMMAPI implements AMMAPI {
     private async _getStableLpTokens(): Promise<Array<StableLpToken>> {
         const stablePools = await this.api.query.zenlinkStableAmm.pools.entries();
         const stableLpTokens = stablePools.map(([key, poolData]) => {
+            if (!poolData.isSome) {
+                return null;
+            }
             const poolBase = DefaultAMMAPI.getStablePoolInfo(poolData.unwrap());
             if (poolBase === null) {
                 return null;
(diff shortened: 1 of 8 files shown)


src/parachain/amm.ts
```diff
@@ -176,7 +198,7 @@ export class DefaultAMMAPI implements AMMAPI {
     private async _getStableLpTokens(): Promise<Array<StableLpToken>> {
         const stablePools = await this.api.query.zenlinkStableAmm.pools.entries();
         const stableLpTokens = stablePools.map(([key, poolData]) => {
-            const poolBase = DefaultAMMAPI.getStableBasePool(poolData.unwrap());
+            const poolBase = DefaultAMMAPI.getStablePoolInfo(poolData.unwrap());
```
bvotteler, line 201: I know this is pre-existing, but noticed the unwrap before checking.
Are we sure that we don't need to filter for `poolData.isSome` ?

src/parachain/amm.ts
```diff
@@ -299,32 +321,126 @@ export class DefaultAMMAPI implements AMMAPI {
...
             // TODO: check number base for fee
-            decodeFixedPointType(poolBase.fee),
+            decodeFixedPointType(poolInfo.fee),
         ];
-
-        const lpToken = DefaultAMMAPI.getStableLpTokenFromPoolData(poolId, poolBase);
-
-        const [pooledCurrencies, apr, A, totalSupply] = await Promise.all([
+        const lpToken = DefaultAMMAPI.getStableLpTokenFromPoolData(poolId, poolInfo);
+        const [pooledCurrenciesBase, apr, A, totalSupply] = await Promise.all([
```
bvotteler, line 350: Maybe a good time to rename `A` to `amplificationCoefficient` or `amplCoeff` or something... not `A` :P

bvotteler (approved):
Can't spot anything obviously wrong. LGTM

src/parachain/amm.ts
peterslany: good idea !

src/parachain/amm.ts
peterslany: thanks, I missed it here :)

PR #554, opened 2023-01-30
chore: add testnet initialization script

sander2:
Todo:
- [ ] relay chain setup
- [ ] amm curve pools


bvotteler (approved):
LGTM

Just one question: Are we okay with all constructed markets pointing to kusama/kintsugi related currencies like KSM, KBTC, and KINT even if this is called with `with-defaults-of="testnet-interlay"` (or `parachain-endpoint` of interlay testnet)?

scripts/init-testnet.ts
```diff
@@ -0,0 +1,318 @@
...
+}
+
+function constructAmmSetup(api: ApiPromise) {
+    const pools = [
+        [{ Token: "KBTC" }, { Token: "KSM" }, 45_000],
+        [{ Token: "KBTC" }, { ForeignAsset: 1 }, 40_000], // usdt
+        [{ Token: "KBTC" }, { ForeignAsset: 2 }, 20_000], // movr
+        [{ Token: "KINT" }, { ForeignAsset: 2 }, 35_000], // movr
+    ];
+    const basicPoolSetup = pools.map(([token1, token2, reward]) => {
```
bvotteler, line 210: You may need to explicitly indicate types for the tuple so TS knows which types to expect.
IIRC, it will think it could be any of the tuple's internal types, ie. treating the input as array of mixed types rather than a tuple.

sander2:
> LGTM
> 
> Just one question: Are we okay with all constructed markets pointing to kusama/kintsugi related currencies like KSM, KBTC, and KINT even if this is called with `with-defaults-of="testnet-interlay"` (or `parachain-endpoint` of interlay testnet)?

Good point, I think I'll leave testnet-interlay for a future pr

scripts/init-testnet.ts
```diff
@@ -0,0 +1,410 @@
...
+            200, // amplification coefficient
+            100_000_000, // max fee 1%
+            0, // no admin fee
+            Buffer.concat([
+                Buffer.from("modl"), // 4 bytes
+                Buffer.from("mod/trsy"), // 8 bytes
+            ], 32), // treasury
+            "(LKSM+VKSM+SKSM)+KSM" // currency symbol
+        ), api.tx.farming.updateRewardSchedule(
+            { StableLpToken: basePoolId + 1 },
```
gregdhill, line 273:
```suggestion
            { StableLpToken: basePoolId + 1 },
```

gregdhill, line 273:
```suggestion
            { StableLpToken: basePoolId + 2 },
```

scripts/init-testnet.ts
```diff
@@ -0,0 +1,410 @@
...
+            { Token: "KINT" },
+            60 * 24 * 7 * 12, // three months, reward period is per minute
+            new BN(10).pow(new BN(12)).muln(15_000),
+        )
+    ];
+
+    const metaPoolSetup = [
+        api.tx.zenlinkStableAmm.createMetaPool(
+            [
+                { StableLpToken: basePoolId + 1}, // LKSM+VKSM+SKSM
```
gregdhill, line 260:
```suggestion
                { StableLpToken: basePoolId }, // LKSM+VKSM+SKSM
```

scripts/init-testnet.ts
```diff
@@ -373,11 +373,11 @@ async function constructAmmSetup(api: ApiPromise) {
         ),
         api.tx.utility.dispatchAs(
             { system: { Signed: treasuryAccount } },
-            api.tx.zenlinkStableAmm.addLiquidity(
+            api.tx.dexStable.addLiquidity(
                 metaPoolId,
                 [
-                    "10_000_000_000_000", // 10 KSM
```
sander2: `"10_000_000_000_000"` gets interpreted as `140004000400040000`

scripts/init-testnet.ts
```diff
@@ -265,36 +265,23 @@ function constructAnnuitySetup(api: ApiPromise) {
...
     decimals.set(JSON.stringify({ Token: "KINT" }), 12);
     decimals.set(JSON.stringify({ ForeignAsset: 1 }), 6); // usdt
     decimals.set(JSON.stringify({ ForeignAsset: 2 }), 18); // movr
-    
+
     const pools: [{ Token: any; }, { Token: any; } | { ForeignAsset: any; }, number, number][] = [
         [
-            { Token: "KBTC" },
             { Token: "KSM" },
+            { Token: "KBTC" },
```
sander2, line 284: Are you sure this is the only one that need to change? Also, this not only impacts `updateRewardSchedule`, but also `createPair` and `addLiquidity`. Won't that create a problem?

gregdhill: Yes the index of the `Token` variant will come before `ForeignAsset`

scripts/init-testnet.ts
```diff
@@ -484,15 +487,15 @@ function constructForeignAssetSetup(api: ApiPromise) {
...
+                    V3: {
                         parents: 1,
                         interior: {
                             X2: [
                                 {
                                     Parachain: 2000
                                 },
                                 {
-                                    GeneralKey: "0x0083"
+                                    GeneralKey: "0x008300000000000000000000000000000000000000000000000000000000000000"
```
sander2, line 498: I'm not sure about this - it seems to me the length is incorrect. But if I remove two zeroes, I get
```
 1:: Enum(GeneralKey):: decodeU8aStruct: failed at 0x83000000000000000000000000000000… on data (index 2/2): [u8;32]:: Expected input with 32 bytes (256 bits), found 31 bytes
```

Issue #550, opened 2023-01-25
AMM: Fix price impact computation for stable pools


peterslany:
Computation of price impact for stable pools is incorrect because it uses `getStableSwapOutputAmount` which is used also in computation of trade execution price -> therefore price impact of trade going through stable pool is always 0.

PR #555, opened 2023-01-31
fix(amm): use price impact computation for stable pools based on curve-js

peterslany:
Fixing computation of price impact going through stable pools. Computation is based on the curve-js method
https://github.com/curvefi/curve-js/blob/master/src/router.ts#L573

Resolves https://github.com/interlay/interbtc-api/issues/550

### DIFF
--- a/src/parachain/amm/trade/utils.ts
+++ b/src/parachain/amm/trade/utils.ts
@@ -1,9 +1,36 @@
 import { CurrencyExt } from "../../../types";
-import { isCurrencyEqual } from "../../../utils";
+import { isCurrencyEqual, newMonetaryAmount } from "../../../utils";
 import { MonetaryAmount } from "@interlay/monetary-js";
 import Big from "big.js";
 import { getStableSwapOutputAmount } from "../liquidity-pool/utils";
-import { isStableMultiPathElement, MultiPath } from "./types";
+import { isStableMultiPathElement, MultiPath, MultiPathElementStable } from "./types";
+
+// SOURCE: Based on curvefi/curve-js middle price calculation method.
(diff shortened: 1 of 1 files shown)


src/parachain/amm/trade/utils.ts
```diff
@@ -1,9 +1,36 @@
...
+// SOURCE: Based on curvefi/curve-js middle price calculation method.
+// https://github.com/curvefi/curve-js/blob/master/src/router.ts#L573
+// @note Will always return 0% price impact for all amounts lower than 0.001
+//       in the same way as curve-js implementation.
+const computeStablePoolMiddlePrice = (
+    currentInputAmount: MonetaryAmount<CurrencyExt>,
+    pathElement: MultiPathElementStable
+): [Big, MonetaryAmount<CurrencyExt>] => {
+    const _getSmallAmountPrice = (amount: MonetaryAmount<CurrencyExt>) => {
+        const decimalsToUse = amount.currency.decimals > 5 ? amount.currency.decimals - 3 : amount.currency.decimals;
```
bvotteler, line 17: I don't quite understand the decimal cutoff to be decimals minus 3 only if we have more than 5 decimals, otherwise the exact decimals.
It has this weird cutover point where 6 decimals get reduced to 3, but 5 decimals stays at 5, and 4 at 4. 

I suppose it's fine as far copy pasta code goes, so no reason for me to reject the PR. Just something we need to be aware of if some results seem odd.

bvotteler (approved):

src/parachain/amm/trade/utils.ts
peterslany: Yes, that seemed strange to me. I think there is also better way to do this, but it would have to be more complex solution that we can use if this one causes issues :+1:

PR #558, opened 2023-02-03
fix(amm): correct decoding of fees and small cleanups

peterslany:
Fixing fee decoding and improving stable pool class.

### DIFF
--- a/src/parachain/amm.ts
+++ b/src/parachain/amm.ts
@@ -43,7 +43,9 @@ import {
 } from "..";
 import { StableLiquidityMetaPool } from "./amm/liquidity-pool/stable-meta";
 
-const HOP_LIMIT = 4; // TODO: add as parameter?
+const HOP_LIMIT = 4;
+const FEE_MULTIPLIER_STANDARD = 1000;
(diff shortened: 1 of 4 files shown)


src/parachain/amm.ts
```diff
@@ -43,7 +43,9 @@ import {
 } from "..";
 import { StableLiquidityMetaPool } from "./amm/liquidity-pool/stable-meta";
 
-const HOP_LIMIT = 4; // TODO: add as parameter?
+const HOP_LIMIT = 4;
+const FEE_MULTIPLIER_STANDARD = 1000;
```
bvotteler, line 47: If we have a way to get these consts from the parachain, can we please add that as future issue/task to github?
And if not... let's chat with backend devs to see if that can be exposed in the future.

Either way: Not blocking.

bvotteler (approved):
LGTM

src/parachain/amm.ts
peterslany: Good point, currently these are not exposed, I'm not sure if it's simple to do that or not though, let's chat with the backend guys!

PR #561, opened 2023-02-08
Peter/amm add farming

peterslany:
Implementing getters for information about pending rewards, farming APR and extrinsic wrapper to claim all pending rewards from farming.


bvotteler (approved):
LGTM

PR #569, opened 2023-02-22
script: verification of vault capacity migration

sander2:
This script is quite ugly in places but I didn't want to spend time on making it prettier. Committing because it might be useful in hacking together future scripts

### DIFF
--- a/scripts/verify-vault-capacity.ts
+++ b/scripts/verify-vault-capacity.ts
@@ -1,0 +1,71 @@
+import { ApiPromise, WsProvider } from '@polkadot/api';
+import { isNull } from '@polkadot/util';
+import { BN } from "bn.js";
+
+async function main() {
+    const wsProvider = new WsProvider('wss://api-kusama.interlay.io/parachain');
+    const api = await ApiPromise.create({ provider: wsProvider });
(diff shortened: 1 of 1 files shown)


scripts/verify-vault-capacity.ts
```diff
@@ -0,0 +1,71 @@
...
+        let isActive = JSON.stringify(vaultJSON.status).includes('true');
+        
+        // inactive vaults don't contribute capacity
+        if (!isActive) {
+            continue;
+        }
+        let n = JSON.stringify(vaultJSON.secureCollateralThreshold).includes("null");
+
+        let vaultId = vaultJSON.id;
+        let collateral = new BN((await apiAt.query.vaultStaking.totalCurrentStake(0, vaultId)).toString());
```
daniel-savu, line 41: I can't see where the capacity pool stake is queried, is it this one?

sander2: Capacity pool stake is now queried at all. The script outputs the _expected_ capacity stakes, and then you need to manually compare that to the migrated storage

daniel-savu (approved):

Issue #567, opened 2023-02-16
Query undercollateralized borrowers


daniel-savu:
Add a function `getUndercollateralizedBorrowers(...)` that computes undercollateralized borrowers, using the `loans.getLiquidationThresholdLiquidity` RPC call, and integration test it. Possibly also add a subscription version of this function.

PR #570, opened 2023-02-22
feat(loans): get undercollateralized borrowers

daniel-savu:
Closes https://github.com/interlay/interbtc-api/issues/567

Breaking changes:
- the `currency` property is removed from `LoanPosition` because it's not required
- `LendPosition` is renamed to `CollateralPosition` because it's easier to distinguish from `LoanPosiiton`
- Unrelated, but since we're make breaking changes: `getBlockRewardAPY` and `computeReward` have their unused parameters removed

### DIFF
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -4,12 +4,12 @@ import {
     BorrowPosition,
     CurrencyExt,
     LoanAsset,
-    LendPosition,
     TickerToData,
     LendToken,
     LoanPosition,
     AccountLiquidity,
     WrappedCurrency,
+    CollateralPosition,
 } from "../types";
 import { AssetRegistryAPI } from "./asset-registry";
 import { ApiPromise } from "@polkadot/api";
@@ -39,7 +39,7 @@ export interface LoansAPI {
      * @param accountId the account Id for which to get supply positions
      * @returns Array of lend positions of account.
      */
-    getLendPositionsOfAccount(accountId: AccountId): Promise<Array<LendPosition>>;
+    getLendPositionsOfAccount(accountId: AccountId): Promise<Array<CollateralPosition>>;
 
     /**
      * Get the borrow positions for given account.
@@ -317,7 +317,7 @@ export class DefaultLoansAPI implements LoansAPI {
         underlyingCurrency: CurrencyExt,
         underlyingCurrencyId: InterbtcPrimitivesCurrencyId,
         lendTokenId: InterbtcPrimitivesCurrencyId
-    ): Promise<LendPosition | null> {
+    ): Promise<CollateralPosition | null> {
         const [underlyingCurrencyAmount] = await this.getLendPositionAmounts(
             accountId,
             lendTokenId,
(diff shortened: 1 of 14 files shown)


src/types/loans.ts
```diff
@@ -38,4 +38,9 @@ interface LoanMarket {
     lendTokenId: number;
 }
 
-export type { LoanPosition, LendPosition, BorrowPosition, LoanAsset, TickerToData, LoanMarket };
+type AccountLiquidity = {
+    liquidity: Big;
```
daniel-savu, line 42: Would in fact be cleaner if this were a `MonetaryAmount<WrappedCurrency>`

src/parachain/loans.ts
```diff
@@ -170,11 +182,22 @@ export interface LoansAPI {
         repayAmount: MonetaryAmount<CurrencyExt>,
         collateralCurrency: CurrencyExt
     ): Promise<void>;
+    /**
+     * @returns An array of `[AccountId, Shortage]` tuples, where `Shortage` is expressed in the
```
daniel-savu, line 186: Need to modify this doc comment still

src/parachain/loans.ts
```diff
@@ -268,12 +291,48 @@ export class DefaultLoansAPI implements LoansAPI {
...
+                shortfall: liquidity[i].shortfall,
+                collateralPositions: collateral[i],
+                borrowPositions: borrows[i],
+            });
+        }
+        return undercollateralizedPositions.filter((position) => !position.shortfall.isZero());
+    }
+
+    async getLiquidationThresholdLiquidity(accountId: AccountId): Promise<AccountLiquidity> {
+        const [rawLiquidity, rawShortfall] = await this.api.rpc.loans.getLiquidationThresholdLiquidity(accountId);
```
bvotteler, line 323: Does this always return a value for any account?

test/utils/helpers.ts
```diff
@@ -381,3 +381,19 @@ export function getCorrespondingCollateralCurrenciesForTests(
...
+ * @param arr
+ * @param element
+ * @returns a boolean representing the presence of the stringified version of `elements` amongts the
+ * stringified items of `arr`
+ */
+export function includesStringified<T extends ImplementsToString>(arr: Array<T>, element: T): boolean {
+    if (arr.length == 0) {
+        return false;
+    }
+    return arr.map((accountId) => accountId.toString() == element.toString()).reduce((acc, value) => acc || value);
```
bvotteler, line 398: I see where the name `accountId` originated from, but maybe we want to rename it after adding the more abstract method. :)

bvotteler (approved):
LGTM

src/parachain/loans.ts
daniel-savu: Good point - yes. I added this getter to the interface so I can document it, also added some doc comments to the `AccountLiquidity` type.

test/utils/helpers.ts
daniel-savu: Good catch :D

PR #571, opened 2023-02-23
Peter/loans feat fetch prices on chain

peterslany:
Moving price fetching from UI to lib, replacing CoinGecko and using on-chain oracle as price source for LTV, collateralization and average thresholds computation.

Added `getCollateralInfo` method takes parameters `lendPositions`, `borrowPositions`, and `loanAssets`. All of these are returned from lib methods `getLendPositionsOfAccount`, `getBorrowPositionsOfAccount` and `getLoanAssets`, respectively. `getLoanAssets` was extended with call to oracle pallet to get asset's exchange rate against bitcoin which is exposed in the `LoanAsset` object. Therefore `getCollateralInfo` method can be run synchronously. This method computes and returns `LoanCollateralInfo`.
`LoanCollateralInfo` is interface consisting of following properties and methods that can be later used on UI side:
- `totalLentBtc` - total aggregated amount of all supplied assets in BTC value
- `totalCollateralBtc` - total aggregated amount of all supplied assets used as collateral in BTC value
- `totalBorrowedBtc` - total aggregated amount of all borrowed assets in BTC value
- `borrowLimitBtc` - total amount that can be borrowed until collateral threshold is met, denominated in BTC
- `ltv` - loan to value 
- `collateralThresholdWeightedAverage` - Average collateral threshold. Denominated in decimals. 
- `liquidationThresholdWeightedAverage` - Average liquidation threshold. Denominated in decimals.
- `calculateBorrowLimitBtcChange` - helper function to compute values based on user's input - to replace similar helper on UI
- `calculateLtvAndThresholdsChange` -helper function that computes new LTV and thresholds based on user's action and input. Replacement for helper on UI.

Note:
All of these properties and methods are equivalent to what is being computed on UI side and should be used to replace UI logic. 

**Next steps on the UI side:**
- Replace logic on UI side with call to `getCollateralInfo` which already pre-computes all the needed values.
- Convert BTC values to USD using price from CoinGecko - should be only for preview purposes

### DIFF
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -375,14 +375,14 @@ export class DefaultLoansAPI implements LoansAPI {
         loanAssets: TickerToData<LoanAsset>
     ): LoanCollateralInfo {
         const lendCollateralPositions = lendPositions.filter(({ isCollateral }) => isCollateral);
-        const lendCollateralThresholdAdjustedPositions = lendPositions.map((position) => {
+        const lendCollateralThresholdAdjustedPositions = lendCollateralPositions.map((position) => {
             const collateralTheshold = loanAssets[position.currency.ticker].collateralThreshold;
             return {
                 ...position,
                 amount: adjustToThreshold(position.amount, collateralTheshold),
             };
         });
-        const lendLiquidationThresholdAdjustedPositions = lendPositions.map((position) => {
+        const lendLiquidationThresholdAdjustedPositions = lendCollateralPositions.map((position) => {
             const liquidationThreshold = loanAssets[position.currency.ticker].liquidationThreshold;
             return {
                 ...position,
@@ -544,8 +544,8 @@ export class DefaultLoansAPI implements LoansAPI {
         const wrappedCurrencyId = this.api.consts.escrowRewards.getWrappedCurrencyId;
         const wrappedCurrency = tokenSymbolToCurrency(wrappedCurrencyId.asToken);
         if (isCurrencyEqual(fromCurrency, wrappedCurrency)) {
-            const oneWrappedCurrencyAtomic = Big(10).pow(wrappedCurrency.decimals);
-            return new ExchangeRate(wrappedCurrency, wrappedCurrency, oneWrappedCurrencyAtomic);
+            const wrappedCurrencyToBitcoinRate = Big(1); // Use exchange rate of 1 between wrapped BTC and real BTC.
+            return new ExchangeRate(wrappedCurrency, wrappedCurrency, wrappedCurrencyToBitcoinRate);
         }
         return this.oracleAPI.getExchangeRate(fromCurrency);
     }
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -544,7 +544,7 @@ export class DefaultLoansAPI implements LoansAPI {
         const wrappedCurrencyId = this.api.consts.escrowRewards.getWrappedCurrencyId;
         const wrappedCurrency = tokenSymbolToCurrency(wrappedCurrencyId.asToken);
         if (isCurrencyEqual(fromCurrency, wrappedCurrency)) {
-            const wrappedCurrencyToBitcoinRate = Big(1); // Use exchange rate of 1 between wrapped BTC and real BTC.
+            const wrappedCurrencyToBitcoinRate = Big(1);
             return new ExchangeRate(wrappedCurrency, wrappedCurrency, wrappedCurrencyToBitcoinRate);
         }
         return this.oracleAPI.getExchangeRate(fromCurrency);
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -8,7 +8,7 @@ import {
     TickerToData,
     LendToken,
     LoanPosition,
-    LoanCollateralInfo,
+    LendingStats,
     WrappedCurrency,
 } from "../types";
 import { ApiPromise } from "@polkadot/api";
@@ -64,12 +64,13 @@ export interface LoansAPI {
      * @param borrowPositions Borrow positions of account.
      * @param loanAssets All loan assets data in TickerToData structure.
      * @return Collateral information about account based on passed positions.
+     * @throws When `loanAssets` does not contain all of the loan positions currencies.
      */
-    getLoanCollateralInfo(
+    getLendingStats(
         lendPositions: Array<LendPosition>,
         borrowPositions: Array<BorrowPosition>,
         loanAssets: TickerToData<LoanAsset>
-    ): LoanCollateralInfo;
+    ): LendingStats | undefined;
 
     /**
      * Get all loan assets.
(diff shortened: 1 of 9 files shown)


src/parachain/loans.ts
```diff
@@ -435,18 +540,29 @@ export class DefaultLoansAPI implements LoansAPI {
         return newMonetaryAmount(amount, rewardCurrency);
     }
 
+    async _getExchangeRate(fromCurrency: CurrencyExt): Promise<ExchangeRate<WrappedCurrency, CurrencyExt>> {
+        const wrappedCurrencyId = this.api.consts.escrowRewards.getWrappedCurrencyId;
+        const wrappedCurrency = tokenSymbolToCurrency(wrappedCurrencyId.asToken);
+        if (isCurrencyEqual(fromCurrency, wrappedCurrency)) {
+            const wrappedCurrencyToBitcoinRate = Big(1); // Use exchange rate of 1 between wrapped BTC and real BTC.
```
daniel-savu, line 547: > // Use exchange rate of 1 between wrapped BTC and real BTC.

As far as I can see, `isCurrencyEqual` checks for ticker equality. In which case would `fromCurrency` or `wrappedCurrency` be actual BTC?

src/types/loans.ts
```diff
@@ -15,6 +15,22 @@ interface BorrowPosition extends LoanPosition {
     accumulatedDebt: MonetaryAmount<CurrencyExt>;
 }
 
+type LoanAction = "lend" | "withdraw" | "borrow" | "repay";
```
daniel-savu, line 18: What's the overall approach to using this `LoanAction` type? I can see that the amounts in `LoanCollateralInfo` are updated based on it, is that meant to just show quick visual feedback after a tx is sent? For instance, if a "liquidation" action occurs (or if the user withdraws/borrows etc with a script), those changes wouldn't be caught here

src/types/loans.ts
```diff
@@ -15,6 +15,22 @@ interface BorrowPosition extends LoanPosition {
...
+interface LoanCollateralInfo {
```
daniel-savu, line 19: This interface contains information about a lot of aspects of the lending protocol, and `LoanCollateralInfo` is misleasing. I'd rename more generically to just `LoansStatus` or similar.

src/parachain/loans.ts
```diff
@@ -335,6 +369,77 @@ export class DefaultLoansAPI implements LoansAPI {
         return this._getPositionsOfAccount(accountId, this._getBorrowPosition.bind(this));
     }
 
+    getLoanCollateralInfo(
```
daniel-savu, line 372: Also rename here to match the new name of the return type

src/parachain/loans.ts
```diff
@@ -435,18 +540,29 @@ export class DefaultLoansAPI implements LoansAPI {
...
         const underlyingCurrency = await currencyIdToMonetaryCurrency(this.api, underlyingCurrencyId);
 
-        const [lendApy, borrowApy, [totalLiquidity, availableCapacity, totalBorrows], rewardCurrency] =
+        const [lendApy, borrowApy, [totalLiquidity, availableCapacity, totalBorrows], rewardCurrency, exchangeRate] =
             await Promise.all([
                 this._getLendApy(underlyingCurrencyId),
                 this._getBorrowApy(underlyingCurrencyId),
                 this._getTotalLiquidityCapacityAndBorrows(underlyingCurrency, underlyingCurrencyId),
                 this._getRewardCurrency(),
+                this._getExchangeRate(underlyingCurrency),
```
daniel-savu, line 565: Will this become the only place where oracle prices are fetched in the UI?

test/unit/parachain/loans.test.ts
```diff
@@ -128,4 +134,173 @@ describe("DefaultLoansAPI", () => {
...
+            ).to.be.true;
+        });
+
+        it("should compute correct LTV and average thresholds", () => {
+            const governanceCurrencyLentAmount = Big(100);
+            const wrappedCurrencyLentAmount = Big(2);
+            const relayCurrencyLentAmount = Big(1000); // Should not affect the computation since it won't be enabled as collateral.
+
+            const governanceCurrencyBorrowedAmount = Big(10);
+            const relayCurrencyBorrowedAmount = Big(10);
```
daniel-savu, line 249: Here too, please use `MonetaryAmount` instead of `Big`. The signatures of `mockLendPosition` and `mockBorrowPosition` should be easy to change so you don't have to unroll the object when calling the functions

src/parachain/loans.ts
```diff
@@ -335,6 +369,77 @@ export class DefaultLoansAPI implements LoansAPI {
...
     }
 
+    getLoanCollateralInfo(
+        lendPositions: Array<LendPosition>,
+        borrowPositions: Array<BorrowPosition>,
+        loanAssets: TickerToData<LoanAsset>
+    ): LoanCollateralInfo {
+        const lendCollateralPositions = lendPositions.filter(({ isCollateral }) => isCollateral);
+        const lendCollateralThresholdAdjustedPositions = lendCollateralPositions.map((position) => {
+            const collateralTheshold = loanAssets[position.currency.ticker].collateralThreshold;
```
daniel-savu, line 379: > loanAssets[position.currency.ticker]

There's a chance this will throw an error if the currency ticker changes (more likely for foreign assets). It's unlikely, but you should add a doc comment about it at least, so the UI wraps this in a try/catch.

daniel-savu: This can also occur if the `loanAssets` parameter is empty (for whatever reason) while `lendCollateralPositions` is non-empty

test/unit/parachain/loans.test.ts
```diff
@@ -128,4 +134,173 @@ describe("DefaultLoansAPI", () => {
...
+        });
+
+        it("should not throw when there are no positions", () => {
+            expect(loansApi.getLoanCollateralInfo([], [], loanAssets)).to.not.throw;
+        });
+
+        it("should not throw when there are no borrow positions", () => {
+            const lendPositions = [mockLendPosition(testGovernanceCurrency, 1)];
+            expect(loansApi.getLoanCollateralInfo(lendPositions, [], loanAssets)).to.not.throw;
+        });
```
daniel-savu, line 304: You can add another test that passes empty a non-empty `lendPositions`, a non-empty `borrowPositions`, and an empty `TickerToData` for `loanAssets` to make sure it doesn't throw

src/parachain/loans.ts
peterslany: Good point, the comment was misleading, will remove it. I meant that exchange rate is expressed against BTC and in case of wrapped BTC, the protocol assumes it has the exactly same value as BTC, right?

src/types/loans.ts
peterslany: This type union is useful when UI needs to compute how the LTV metrics will change once `LoanAction` is done - before submitting the tx. It is used as in  `calculateBorrowLimitBtcChange`, `calculateLtvAndThresholdsChange`.

src/types/loans.ts
peterslany: Naming this one is not easy, `LoansStatus` is already used on UI to denote risk of getting liquidated, so I would use `LoansAccountInfo`, what do you think about that one? Open to other suggestions!

src/parachain/loans.ts
peterslany: No, the oracle price is used fetched in other places in lib that are used in UI too - vault capacity calculation, vault apy calculation.

src/parachain/loans.ts
daniel-savu: >  the protocol assumes it has the exactly same value as BTC, right?

It does yeah

src/types/loans.ts
daniel-savu: `LendingStats`?

src/parachain/loans.ts
daniel-savu: For now it's fine but the UI could optimize this by fetching exchange rates just once and then passing them down to the various components, the same way `loanAssets` is passed in `getLoanCollateralInfo`. If you think this makes sense you can create a GitHub issue in the UI/lib.

src/types/loans.ts
peterslany: sounds good :+1:

src/parachain/loans.ts
peterslany: Actually, this function has to throw if `loanAssets` object does not contain data for some of the currencies that loan positions consist of. I added function to check this and return meaningful error message if such case occurs.

test/unit/parachain/loans.test.ts
peterslany: Added test to ensure that it _throws_, see my response to the previous comment.

src/parachain/loans.ts
peterslany: I think it's not necessary at the moment since in the other place the exchange rate is fetched only for one currency - so fetching all extra exchange rates for loan currencies would not make the UI more efficient.

peterslany:
@daniel-savu I edited the PR with your comments, please re-review the PR when you are back :pray:

PR #573, opened 2023-02-23
chore: set all versions in runtime-upgrade script

gregdhill:
Shamelessly taken from https://github.com/interlay/interbtc-api/pull/554

### DIFF
--- a/scripts/runtime-upgrade.ts
+++ b/scripts/runtime-upgrade.ts
@@ -90,7 +90,7 @@ async function setAllClientReleases(api: ApiPromise, baseUrl: String, runtimeNam
     }
 
     return matches.map(([checksum, fullFileName, clientName]) => {
-        return api.tx.clientsInfo.setCurrentClientRelease(
+        return api.tx.clientsInfo.setPendingClientRelease(
             clientName,
             {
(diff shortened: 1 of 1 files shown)


scripts/runtime-upgrade.ts
```diff
@@ -73,6 +73,33 @@ async function printDiscordProposal(
...
+async function setAllClientReleases(api: ApiPromise, baseUrl: String, runtimeName: String) {
+    const checksumFile = await fetch(baseUrl + 'sha256sums.txt')
+        .then(res => {
+            if (res.status >= 400) {
+                throw new Error("Bad response from server");
+            }
+            return res.text();
+        });
+
+    const regex = new RegExp("([a-f0-9]+)\\\s*[.]\/(([a-z]+)-parachain-metadata-" + runtimeName + ")", "g");
```
sander2, line 85: i think this is matching both the testnet and mainnet binaries

sander2 (changes requested):
We probably shouldn't set faucet url on mainnets

sander2 (commented):
It should set the pending release rather than the current one

sander2 (approved):

PR #574, opened 2023-02-27
feat(scripts): see undercollateralized borrowers in CLI

daniel-savu:
- Adds script for querying undercollateralized borrowers. Results are sorted in descending order by shortfall and the various collateral / borrow positions are separated by `|` and rounded down to three decimals.
- Adds `disconnect()` method on the API to avoid scripts from hanging when they finish execution
- Silences `polkadot-js/api` warnings by default (such as the ones regarding missing RPC decorations)

Example result when running `$ yarn undercollateralized-borrowers --parachain-endpoint wss://api-dev-kintsugi.interlay.io/parachain`

Screenshot from 2023-02-27 14-31-40


src/interbtc-api.ts
```diff
@@ -203,4 +204,8 @@ export class DefaultInterBtcApi implements InterBtcApi {
         // beware: this call will throw if the wrapped currency is not a token!
         return tokenSymbolToCurrency(currencyId.asToken);
     }
+
+    public disconnect(): void {
+        this.api.disconnect();
```
bvotteler, line 209: I'd suggest adding this as `Promise<void>`, given `ApiPromise.disconnect()` is async.

bvotteler (approved):
LGTM

src/interbtc-api.ts
daniel-savu: That what my brain wanted to do but somehow the function turned out this way haha

PR #579, opened 2023-02-28
chore: don't test redeem fee bumping

gregdhill:
Opening this PR to discuss the merits of removing / keeping these tests. We anyway test this functionality in the clients repository so I'm not sure it is worth duplicating that here.

### DIFF
--- a/test/integration/parachain/staging/sequential/redeem.test.ts
+++ b/test/integration/parachain/staging/sequential/redeem.test.ts
@@ -3,7 +3,6 @@ import { KeyringPair } from "@polkadot/keyring/types";
 import { Hash } from "@polkadot/types/interfaces";
 import {
     AssetRegistryAPI,
-    currencyIdToMonetaryCurrency,
     DefaultAssetRegistryAPI,
     DefaultInterBtcApi,
     DefaultLoansAPI,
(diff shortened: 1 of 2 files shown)


bvotteler:
> We anyway test this functionality in the clients repository so I'm not sure it is worth duplicating that here.

That's great to hear! 
I remember we added this back in the day because the tests weren't complete in the clients, or missed some cases. If we're happy now, we can drop them.

Pinging @sander2 for feedback - I was working on this with his guidance wrt testing the gaps.

sander2:
Can you link the specific test(s) you mean in the client? If we indeed cover this fully then I'm fine with removing

gregdhill:
@sander2 https://github.com/interlay/interbtc-clients/blob/6b30c7c51783d260fefa03ec7dabcc06921f9105/vault/tests/vault_integration_tests.rs#L1106

sander2:
The client test seems indeed sufficient to me. I checked also that it actually runs in the pipeline (since it's feature guarded). I'm not sure why we initially went for testing in the lib - possibly this was before we had bitcoind integration tests in the client

bvotteler (approved):
LGTM with Sander's blessing

PR #581, opened 2023-03-01
chore: remove all unused code

gregdhill:
Clean up constants, remove `set-state` and util setup script, fix fee estimation type


bvotteler (approved):
LGTM

Issue #585, opened 2023-03-01
Cleanup public interfaces


gregdhill:
There are many methods exposed on the public interfaces such as the staking methods on the `RewardsAPI` that can be removed since they have no use downstream.

PR #586, opened 2023-03-01
chore: remove unused rewards methods

gregdhill:
Unused, related: https://github.com/interlay/interbtc-api/issues/585


daniel-savu:
> Checking in with @daniel-savu : Do the bots use any of these removed APIs?
> 
> (I checked the UI repo, and they're not used there.)

No dependencies on these APIs from the bots, so can be safely removed

bvotteler (approved):
LGTM, thank you!

PR #588, opened 2023-03-01
refactor: use api for storage setters

gregdhill:
Use polkadot-js for storage key / value encoding instead of rolling it ourselves


bvotteler (approved):
Awesome stuff, thank you.

bvotteler (approved):

PR #597, opened 2023-03-06
chore: remove now unused staking getters

gregdhill:
Not used in the UI so not breaking, but I would like for us to remove the `RewardsAPI` at some point


bvotteler (approved):
LGTM

PR #600, opened 2023-03-07
test: unskip multi-vault issue batching test

gregdhill:
Seems to already work without changes, not sure why it was skipped before. Left the "FIXME" note since I'm not sure the test would be improved by issuing exactly the allowance of two vaults.


bvotteler (approved):
LGTM

PR #603, opened 2023-03-07
test: should request vault replacement

gregdhill:
Final remaining "skipped" test from the before times, also adds some params to the replace api but since they are optional it shouldn't break the interface

### DIFF
--- a/test/integration/parachain/staging/sequential/replace.test.ts
+++ b/test/integration/parachain/staging/sequential/replace.test.ts
@@ -10,6 +10,8 @@ import {
     InterbtcPrimitivesVaultId,
     LoansAPI,
     newMonetaryAmount,
+    sleep,
+    SLEEP_TIME_MS,
 } from "../../../../../src/index";
 
 import { BitcoinCoreClient } from "../../../../../src/utils/bitcoin-core-client";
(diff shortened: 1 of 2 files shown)


test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -155,9 +180,8 @@ describe("replace", () => {
                 // make sure vault does not hold enough issued tokens to request a replace
                 const replaceAmount = dustValue.add(tokensInVault);
 
-                const replacePromise = callWith(interBtcAPI, vault_2, () =>
-                    interBtcAPI.replace.request(replaceAmount, collateralCurrency)
-                );
+                interBtcAPI.setAccount(vault_2);
```
bvotteler, line 183: I believe we can move these introduced `setAccounts` outside the loop. Not a huge boost, though, so not critical imo.

test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -103,38 +110,56 @@ describe("replace", () => {
                     loansAPI,
                     vault_3_id.currencies.collateral
                 );
-                await callWith(interBtcAPI, vault_3, async () =>
-                    interBtcAPI.replace.request(replaceAmount, collateralCurrency)
+
+                console.log(`Requesting vault replacement for ${replaceAmount.toString()}`);
+                interBtcAPI.setAccount(vault_3);
```
bvotteler, line 115: Same as the other instance: Can be moved outside of the loop I think.

bvotteler (approved):
LGTM

Issue #599, opened 2023-03-07
Pass blockHash to parseRequest methods


gregdhill:
The methods that call this function read only from the finalized head but when parsing the requests we check the latest unfinalized state, this shouldn't be an issue in practice but it is better to be consistent and read from the same height.

PR #604, opened 2023-03-08
Feat AMM

peterslany:
Merge AMM branch back into master.


src/utils/loans.ts
```diff
@@ -0,0 +1,159 @@
...
+        case "borrow":
+            return currentTotalBorrowedBtc.add(actionAmountBtc);
+        case "repay":
+            return currentTotalBorrowedBtc.sub(actionAmountBtc);
+        default:
+            return currentTotalBorrowedBtc;
+    }
+};
+
+const calculateCollateralAmountBtc = (
```
bvotteler, line 21: Naming between this method and `calculateTotalBorrowedBtcChange` (ends with "Change") above seems to be inconsistent. I might be missing something, though.

test/unit/parachain/loans.test.ts
```diff
@@ -161,4 +134,181 @@ describe("DefaultLoansAPI", () => {
...
+                `Average collateral threshold: ${collateralThresholdWeightedAverage.toString()} does not match expected threshold: ${expectedAverageCollateralThreshold.toString()}`
+            ).to.be.true;
+            expect(
+                liquidationThresholdWeightedAverage.eq(expectedAverageLiquidationThreshold),
+                `Average liquidation threshold: ${liquidationThresholdWeightedAverage.toString()} does not match expected threshold: ${expectedAverageLiquidationThreshold.toString()}`
+            ).to.be.true;
+        });
+
+        it("should not throw when there are no positions", () => {
+            expect(() => loansApi.getLendingStats([], [], loanAssets)).to.not.throw;
```
bvotteler, line 300: Not sure about this, but do we need `await`s here to avoid promises being still evaluated after this test suite is marked as complete?

test/unit/utils/rewards.test.ts
```diff
@@ -0,0 +1,34 @@
...
+            const actualRewardAmount = calculateAnnualizedRewardAmount(amountPerBlock, assumedBlockDurationSecs * 1000);
+
+            expect(actualRewardAmount.eq(expectedRewardAmount)).to.eq(
+                true,
+                `Expected result to be equal to ${expectedRewardAmount.toString()} but was: ${actualRewardAmount.toString()}`
+            );
+        });
+
+        it("should throw if blockTimeMs is zero", () => {
+            expect(() => calculateAnnualizedRewardAmount(Big(1), 0)).to.throw("Division by zero");
```
bvotteler, line 25: Similar here, I suspect we might need an `await` here, so the throw doesn't happen outside of the `describe` block if it occurs. (Not 100% sure tho)

bvotteler (commented):
Looking good for the most part. Left comments on the points needing clarification.

Expecting quite a few changes around the tests in particular once master is merged back in.

src/parachain/loans.ts
```diff
@@ -348,7 +349,7 @@ export class DefaultLoansAPI implements LoansAPI {
         if (underlyingCurrencyAmount.eq(0)) {
             return null;
         }
-
+        const amount = newMonetaryAmount(underlyingCurrencyAmount, underlyingCurrency);
```
gregdhill, line 352: `amount` is unused

src/parachain/amm.ts
```diff
@@ -0,0 +1,925 @@
...
+            deadline
+        );
+
+        return [withdrawLiquidityExtrinsic, this.api.events.dexStable.RemoveLiquidity];
+    }
+
+    async removeLiquidity(
+        amount: MonetaryAmount<LpCurrency>,
+        pool: LiquidityPool,
+        maxSlippage: number, // Percentage.
```
gregdhill, line 869: Won't block on this, but just wondering if we could type "Percentage" to make it explicit what value should be passed?

src/utils/currency.ts
```diff
@@ -143,22 +170,34 @@ export function getCurrencyIdentifier(currency: CurrencyExt): CurrencyIdentifier
...
+        const token1 = getCurrencyIdentifier(currency.lpToken.token1) as StandardPooledTokenIdentifier;
+        return { lpToken: [token0, token1] };
+    }
+
     return { token: currency.ticker };
 }
 
 export async function currencyIdToMonetaryCurrency(
-    assetRegistryApi: AssetRegistryAPI,
-    loansApi: LoansAPI,
```
gregdhill, line 151: This is a nice change, should be easier to close this issue now: https://github.com/interlay/interbtc-api/issues/599

gregdhill (commented):
We should be able to remove most of the path finding code once this is merged: https://github.com/interlay/interbtc/pull/944

scripts/docker-setup.sh
```diff
@@ -23,7 +23,7 @@ if ! [ -d "local-setup" ]
 then
     mkdir local-setup
     git clone https://github.com/interlay/parachain-launch/
-    cd parachain-launch && git checkout 1.1.0-p1.21.3-c1.19.1 && yarn install
+    cd parachain-launch && git checkout 1.1.0-p1.21.7-c1.19.1 && yarn install
```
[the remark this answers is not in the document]
peterslany: Yes, this script was already removed form master, so removing from this PR as well :+1:

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interlay/interbtc-api",
-  "version": "1.20.3",
+  "version": "0.32.10",
```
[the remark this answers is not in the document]
peterslany: I used `1.21.0`, I would increase major version only if we do so on parachain side too to stay consistent. what do you think?

src/parachain/amm.ts
```diff
@@ -0,0 +1,925 @@
...
+                .map(
+                    (actuallyPooledCurrency) =>
+                        minAmounts.find((amount) =>
+                            isCurrencyEqual(amount.currency, actuallyPooledCurrency.currency)
+                        ) || new MonetaryAmount(actuallyPooledCurrency.currency, 0)
+                )
+                .map(monetaryAmountToRawString);
+            // const minAmountsMeta = minAmounts
+            //     .filter((amount) => pool.involvesToken(amount.currency))
+            //     .map(monetaryAmountToRawString);
```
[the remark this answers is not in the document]
peterslany: Left over comment, thanks for finding!

src/parachain/amm/liquidity-pool/stable.ts
```diff
@@ -0,0 +1,283 @@
+import { CurrencyExt, StableLpToken } from "../../../types";
+import { isCurrencyEqual } from "../../../utils";
+import { MonetaryAmount } from "@interlay/monetary-js";
+import Big from "big.js";
+import { PoolType, PooledCurrencies } from "../types";
+import { LiquidityPoolBase } from "./types";
+import { LiquidityPoolCalculator } from "./calculator";
+
+// Maximum decimal precision.
+const DECIMAL_PRECISION = Big(10).pow(18);
```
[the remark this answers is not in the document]
peterslany: This constant is used as a precision point when approximating pool invariant variables using Newton's method, I moved it to `constants` file and added more informative comment and name.

src/utils/loans.ts
peterslany: You are right, changed naming to make it consistent :)

test/unit/parachain/loans.test.ts
peterslany: No, in this case there is no need to await, because `getLendingStats` does not return promise.

test/unit/utils/rewards.test.ts
peterslany: This was removed from master, removing from PR.

package.json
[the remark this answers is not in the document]
bvotteler: The way I look at minor version bumps for the lib is: If we have a breaking change in an existing API we definitely bump at least the minor version.

Given these changes are quite big, with whole new groups of APIs available, I think a minor version bump is justified.

And we're not aligned with clients/parachain versions anyway, so I feel we can justify a bit more wriggle room in our choices :)

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@interlay/interbtc-api",
   "version": "1.21.0",
-  "description": "JavaScript library to interact with interBTC",
+  "description": "JavaScript library to interact with interBTC and kBTC",
```
bvotteler, line 4: Paging @gregdhill , I seem to recall you removed this ("and kBTC") in your other PR?

src/parachain/rewards.ts
```diff
@@ -57,59 +55,29 @@ export class DefaultRewardsAPI implements RewardsAPI {
...
-                        nominatorId
-                    ]
-                ]
-            ],
-        ]).then((data) => data.map((value) => decodeFixedPointType(value)));
+        const [stake, slashPerToken, slashTally] = await this.api
+            .queryMulti<[SignedFixedPoint, SignedFixedPoint, SignedFixedPoint]>([
+                [this.api.query.vaultStaking.stake, [nonce, [vaultId, nominatorId]]],
+                [this.api.query.vaultStaking.slashPerToken, [nonce, vaultId]],
+                [this.api.query.vaultStaking.slashTally, [nonce, [vaultId, nominatorId]]],
```
bvotteler, line 64: TIL you can share a nonce across `queryMulti` queries. Nice!

src/utils/constants.ts
```diff
@@ -20,4 +20,7 @@ export const SLEEP_TIME_MS = 1000;
 
 export function sleep(ms: number): Promise<void> {
     return new Promise((resolve) => setTimeout(resolve, ms));
-}
\ No newline at end of file
+}
+
+// Decimal precision of Newton's approximations that are computed in this class.
+export const STABLE_POOLS_APPROXIMATION_PRECISION = Big(10).pow(9);
```
bvotteler, line 26: Wasn't that `.pow(18)` before? Or is this a different value?

src/utils/currency.ts
```diff
@@ -185,3 +224,107 @@ export function tokenSymbolToCurrency(tokenSymbol: InterbtcPrimitivesTokenSymbol
...
+ * @param lendTokenId Currency id of the lend token to get currency from
+ * @returns Underlying CurrencyExt for provided lend token
+ */
+export async function getUnderlyingCurrencyFromLendTokenId(
+    api: ApiPromise,
+    lendTokenId: InterbtcPrimitivesCurrencyId
+): Promise<CurrencyExt> {
+    const underlyingCurrencyId = await api.query.loans.underlyingAssetId(lendTokenId);
+
+    const underlyingCurrency = await currencyIdToMonetaryCurrency(api, underlyingCurrencyId.unwrap());
```
bvotteler, line 266: Do we need an `.isSome` check before unwrapping, or is it guaranteed to be set?

src/utils/rewards.ts
```diff
@@ -0,0 +1,7 @@
+import Big from "big.js";
+import { MS_PER_YEAR } from "./constants";
+
+export function calculateAnnualizedRewardAmount(amountPerBlock: Big, blockTimeMs: number): Big {
```
bvotteler, line 4: Not sure where this is used, but would it make sense to get an rpc for that at some stage?

We've pushed these types of calcs for other scenarios to rpcs - granted some of them were more complex, so it made sense to keep the logic in the parachain code. Just something to consider

bvotteler (commented):
LGTM, just a few smaller things to clarify.

test/utils/helpers.ts
```diff
@@ -210,6 +206,10 @@ export function getCorrespondingCollateralCurrenciesForTests(
     }
 }
 
+export function getWrappedCurrencyForTest(api: ApiPromise): WrappedCurrency {
+    const currencyId = api.consts.currency.getWrappedCurrencyId;
+    return tokenSymbolToCurrency(currencyId.asToken);
+}
```
gregdhill, line 212: Already declared as a getter on `InterBtcApi`

package.json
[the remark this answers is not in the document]
peterslany: Makes sense, I'll tag it as `2.0.0` then :+1:

src/utils/constants.ts
peterslany: It was 18, but I reduced it to 9 to make the computation faster - the approximations are still precise enough for the swap amounts computations.

src/utils/currency.ts
peterslany: Yes, this is always set if the lendToken exists. If not the function will throw - which is correct behaviour for invalid `lendTokenId` provided.

test/utils/helpers.ts
peterslany: Right, I'll reuse that one then :+1:

src/utils/rewards.ts
bvotteler: > Yes, I think adding rpc that would provide calculation with more accurate average block time makes sense!

cc @gregdhill

bvotteler (approved):
LGTM

PR #605, opened 2023-03-08
feat(loans): get loans markets, refactor usage

daniel-savu:
- Adds `getRelayChainCurrency` function
- Exposes `getLoansMarkets` and refactors usage of this function to remove parsing duplication
- Fixes `_getBorrowPosition` bug that fetched borrows using the `lend_token_id` instead of the underlying currency
- Reuses `client-env` anchor in the docker-compose


src/utils/index.ts
```diff
@@ -5,3 +5,4 @@ export * from "./constants";
 export * from "./bitcoin-core-client";
 export * from "./issueRedeem";
 export * from "./storage";
+export * from "./setup";
```
gregdhill, line 8: @daniel-savu this file no longer exists, this is why CI is failing.

src/parachain/oracle.ts
```diff
@@ -93,6 +93,14 @@ export class DefaultOracleAPI implements OracleAPI {
...
 
     async getExchangeRate(currency: CurrencyExt): Promise<ExchangeRate<Bitcoin, CurrencyExt>> {
+        // KBTC / IBTC have an exchange rate of one
+        if (currency.ticker == this.wrappedCurrency.ticker) {
+            return new ExchangeRate<WrappedCurrency, CurrencyExt>(
+                currency,
+                currency,
+                new Big(1),
+            );
+        }
```
gregdhill, line 103: What is the advantage of hardcoding this?

src/parachain/transaction.ts
```diff
@@ -181,6 +181,35 @@ export class DefaultTransactionAPI implements TransactionAPI {
         }
     }
 
+    static async waitForEvent<T extends AnyTuple>(
```
gregdhill, line 184: Remove, not used

src/utils/storage.ts
```diff
@@ -1,8 +1,31 @@
...
+    storageItemName: string,
+    mapItemKey: `0x${string}`,
+    nestedMapItemKey?: `0x${string}`
+): string {
+    let storageKey = getStorageKey(moduleName, storageItemName) + stripHexPrefix(blake2_128Concat(mapItemKey));
+    if (nestedMapItemKey) {
+        storageKey += stripHexPrefix(blake2_128Concat(nestedMapItemKey));
+    }
+    return storageKey;
+}
```
gregdhill, line 28: Remove, use polkadot-js for this instead. See: https://github.com/interlay/interbtc-api/pull/588

gregdhill (changes requested):

src/parachain/oracle.ts
daniel-savu: There is no exchange rate of one for the wrapped currency stored in the aggregate field of the oracle pallet, so it needs to be hardcoded here

src/parachain/transaction.ts
daniel-savu: It's used in the bot, but I can move it there to avoid it being removed from the lib again

src/utils/storage.ts
daniel-savu: `api.query.btcRelay.stableBitcoinConfirmations.key()` <- oh that's very nice!

gregdhill (approved):

peterslany (approved):
Good job, solution looks good, just a minor comment to handle - to remove duplicated code, so giving approval already :)

PR #610, opened 2023-03-10
chore: reduce or remove most extended timeouts

gregdhill:
Catch potential test failures earlier and be explicit about where might take longer


bvotteler (approved):
LGTM, less is more

PR #618, opened 2023-03-21
fix: runtime upgrade script: import fetch and set faucet version

sander2:


scripts/runtime-upgrade.ts
```diff
@@ -3,6 +3,7 @@ import { createSubstrateAPI } from "../src/factory";
 import { ApiPromise } from "@polkadot/api";
 import { cryptoWaitReady, blake2AsHex, sha256AsU8a } from "@polkadot/util-crypto";
 import { SubmittableExtrinsic } from "@polkadot/api/types";
+import fetch from "cross-fetch";
```
daniel-savu, line 6: I assume this was auto-imported when you typed "faucet" but is not really needed?

sander2: > I assume this was auto-imported when you typed "faucet" but is not really needed?

No I explicitly added it since I get `ReferenceError: fetch is not defined` otherwise

daniel-savu (approved):

Issue #612, opened 2023-03-14
AMM: Return pools with no liquidity


peterslany:
Pools with no liquidity are not returned from the lib, but they should be so that everyone can add initial liquidity through the UI.

PR #621, opened 2023-03-27
feat(amm): handle empty pools

peterslany:
Adding `isEmpty` property and returning also empty pools from AMM module.

Closes https://github.com/interlay/interbtc-api/issues/612


bvotteler (approved):
LGTM

Issue #592, opened 2023-03-02
All API methods should return SubmittableExtrinsic


gregdhill:
**Is your feature request related to a problem? Please describe.**
It is difficult to use the current API to compose calls (e.g. in a batch) since we usually return `Promise<void>`.

**Describe the solution you'd like**
The existing methods should not submit the transaction by default but rather return the `SubmittableExtrinsic` and possibly the success event type.

**Describe alternatives you've considered**
Currently we have a bunch of unused "builder" methods which are annoying to maintain.

**Additional context**
The new return type will also allow the consumer to directly monitor the progress of the transaction. This is useful in the frontend for displaying toast popups.

PR #622, opened 2023-03-29
POC refactor to return submittable extrinsic

peterslany:
Opening up POC PR with proposed refactor of how the submission of extrinsics is handled in lib, issue https://github.com/interlay/interbtc-api/issues/592

This refactor introduces `ExtrinsicHandler` class that contains `SubmittableExtrinsic`, optional event type and helper methods for submission, fee estimation and batching. 

I edited AMM module to show how we can easily refactor lib methods to use this approach.

### UI side usage example
**Access `SubmittableExtrinsic` members**
```typescript
const swapExtrinsic = window.bridge.amm.swap(trade);
const submittableExtrinsic = swapExtrinsic.extrinsic;
```
**Submit extrinsic**
```typescript
const swapExtrinsic = window.bridge.amm.swap(trade);
await swapExtrinsic.submit();
```

**Get fee estimation**
```typescript
const swapExtrinsic = window.bridge.amm.swap(trade);
const fee = await swapExtrinsic.getFeeEstimation();
```

**Compose with another extrinsic**
```typescript
// swap and add liquidity
const swapExtrinsic = window.bridge.amm.swap(trade);
const addLiquidityExtrinsic = window.bridge.amm.addLiquidity(params);
const swapAndAddLiquidityExtrinsic = swapExtrinsic.batchWith(addLiquidityExtrinsic);

// submit batched extrinsics
await swapAndAddLiquidityExtrinsic.submit();
```

### DIFF
--- a/src/parachain/amm.ts
+++ b/src/parachain/amm.ts
@@ -47,6 +47,8 @@ import {
     isStandardPool,
     StableLiquidityMetaPool,
 } from "./amm/";
+import { ExtrinsicData } from "../types/extrinsic";
+import { ExtrinsicHandler } from "../utils/extrinsic-handler";
 
 const HOP_LIMIT = 4;
 const FEE_MULTIPLIER_STANDARD = 10000;
@@ -117,7 +119,7 @@ export interface AMMAPI {
         minimumAmountOut: MonetaryAmount<CurrencyExt>,
         recipient: AddressOrPair,
         deadline: number | string
-    ): Promise<void>;
+    ): ExtrinsicHandler;
 
     /**
      * Adds liquidity to liquidity pool
@@ -135,7 +137,7 @@ export interface AMMAPI {
         maxSlippage: number,
         deadline: number,
         recipient: AddressOrPair
-    ): Promise<void>;
+    ): ExtrinsicHandler;
 
     /**
      * Removes liquidity from pool.
@@ -153,7 +155,7 @@ export interface AMMAPI {
         maxSlippage: number,
         deadline: number,
         recipient: AddressOrPair
-    ): Promise<void>;
+    ): ExtrinsicHandler;
 
     /**
      * Claim all pending farming rewards.
@@ -161,7 +163,7 @@ export interface AMMAPI {
      * @param claimableRewards Map of LpToken -> Array of reward monetary amounts -> supposed to be
      *                         output of `getClaimableFarmingRewards`
      */
-    claimFarmingRewards(claimableRewards: Map<LpCurrency, Array<MonetaryAmount<CurrencyExt>>>): Promise<void>;
+    claimFarmingRewards(claimableRewards: Map<LpCurrency, Array<MonetaryAmount<CurrencyExt>>>): ExtrinsicHandler;
 }
 
 export class DefaultAMMAPI implements AMMAPI {
@@ -548,12 +550,12 @@ export class DefaultAMMAPI implements AMMAPI {
         return [...standardPools, ...stablePools];
     }
 
-    private async _swapThroughStandardPoolsOnly(
+    private _swapThroughStandardPoolsOnly(
         trade: Trade,
         minimumAmountOut: MonetaryAmount<CurrencyExt>,
         recipient: AddressOrPair,
         deadline: number | string
-    ): Promise<void> {
+    ): ExtrinsicHandler {
         const { amountIn, amountOutMin, path } = encodeSwapParamsForStandardPoolsOnly(
             this.api,
             trade,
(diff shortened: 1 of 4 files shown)


src/utils/extrinsic-handler.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * This method submits the extrinsic signed by account set in TransactionAPI. Wrapper around TransactionAPI.sendLogged.
+     *
+     * @param {boolean} onlyInBlock Optional parameter to set waiting for transaction inclusion only (true) or finalisation (false).
+     *                              Defaults to true.
+     * @returns {ISubmittableResult} Result of the extrinsic submission.
+     */
+    public submit(onlyInBlock: boolean = true): Promise<ISubmittableResult> {
```
bvotteler, line 44: I wonder if an enum could be more readable here than the boolean?

peterslany: Yes, possibly. How would you name the enum to make it more readable?)

bvotteler: > Yes, possibly. How would you name the enum to make it more readable?)

Maybe a shortened ExtrinsicStatus with only `isInBlock` and `isFinalized` as those are the two options that `DefaultTransactionsApi` looks out for?

(See the full [ExtrinsicStatus definition](https://github.com/polkadot-js/api/blob/4fd3289739e2d0b14eddf7590f729f17f49477cc/packages/types/src/interfaces/author/types.ts#L17-L35) here)

nud3l: I would just make it a pass-through of the `ExtrinsicStatus` and not redefine the states here. The `ExtrinsicStatus` is pretty clearly named and defined and no need to reinvent the wheel on our side.

nud3l (commented):
I'm wondering if we need the wrapper class. By introducing a new class, we require us and others to handle transactions in a special way when using the lib which IMO has two downsides:

1. If polkadot.js adds features or changes, we need logic upstream to handle those changes.
2. Wallets and other parties cannot straightforwardly use our lib. The need to handle the new `ExtrinsicHandler` class instead of just getting the `SubmittableExtrinsic` types for which they might have their own util functions.

I would refactor this PR as follows:

- I can see that the fee estimation method is useful, but I think I would have implemented it rather as a util method on the `transaction API` similar to the `sendLogged` method.
- For `transactionAPI.sendLogged`, I'd remove the `onlyInBlock` boolean and replace it with the `ExtrinsicStatus` as an optional parameter.
- I'd remove the `batchWith` functions. On the transactionAPI, we can have a `batch` function as a helper that just takes in multiple `SubmittableExtrinsic`and provides some extras eventually (like checking for max weight of the added transactions). Removing the `ExtrsinicHandler` also simplifies logic since you just need one `batch` function now.
- I would add a `dryRun(extrinsic: SubmittableExtrinsic) -> bool` function on the `transactionAPI` to give the option to test transactions against the chain before executing them.

src/parachain/amm.ts
```diff
@@ -111,13 +111,14 @@ export interface AMMAPI {
      * @param {MonetaryAmount<CurrencyExt>} minimumAmountOut Minimum output amount to be received.
      * @param {AddressOrPair} recipient Recipient address.
      * @param {number | string} deadline Deadline block for the swap transaction.
+     * @returns {ExtrinsicData} A submittable extrinsic and event.
```
bvotteler, line 114: > A submittable extrinsic and event.

I wonder if this could be a bit confusing?
It returns a submittable extrinsic and the event we expect the parachain to emit when it is submitted / included. I wonder if we should be more explicit about that.

src/parachain/transaction.ts
```diff
@@ -29,10 +33,27 @@ export interface TransactionAPI {
...
     ): SubmittableExtrinsic<"promise", ISubmittableResult>;
+
+    /**
+     * Getter for fee estimation of the extrinsic.
+     *
+     * @param {SubmittableExtrinsic<"promise">} extrinsic Extrinsic to get fee estimation about.
+     * @returns {MonetaryAmount<Currency>} amount of native currency that will be paid as transaction fee.
+     * @note This fee estimation does not include tip.
+     */
+    getFeeEstimation(extrinsic: SubmittableExtrinsic<"promise">): Promise<MonetaryAmount<Currency>>;
```
bvotteler, line 44: nit: I'd probably just call it `getFeeEstimate`
"Estimation" is the process of finding an estimate.

src/parachain/transaction.ts
```diff
@@ -49,12 +70,36 @@ export class DefaultTransactionAPI implements TransactionAPI {
...
+    async dryRun(extrinsic: SubmittableExtrinsic<"promise">): Promise<boolean> {
+        if (this.account === undefined) {
+            return Promise.reject(new Error(ACCOUNT_NOT_SET_ERROR_MESSAGE));
+        }
+        try {
+            await extrinsic.dryRun(this.account);
+            return true;
+        } catch {
+            return false;
+        }
```
bvotteler, line 102: Is there a way to capture / get a preview of the error message?

If so, I think it might make sense to return that in some way.

bvotteler (commented):
Looking great at first glance!

Will take another good look next week.

src/parachain/transaction.ts
peterslany: Ah, nice catch, changed)

src/parachain/transaction.ts
peterslany: Good idea, I changed the return type to `{ success: bool, error?: unknown}` to be able to handle the error later (display in UI, log or so).

test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -103,6 +103,7 @@ describe("replace", () => {
                     interBtcAPI.replace.request(replaceAmount, collateralCurrency),
                     false
                 );
+                console.log("\n\n res, val", result.status.type, result.status.asFinalized);
```
bvotteler, line 106: Is this log still needed? Looks like it might be a leftover from early trials/testing.

bvotteler (approved):
Awesome stuff, thanks!

test/integration/parachain/staging/sequential/replace.test.ts
peterslany: Not needed, good eyes!

PR #627, opened 2023-04-21
Fix: escrow.getRewardEstimate param name change

bvotteler:
Fix misleading parameter name and add better JSDoc.


tomjeatt (approved):
Looks great—much clearer

PR #629, opened 2023-05-09
fix: add incentive reward estimation

peterslany:
Adding more actual incentive reward estimation without user's need to transact with the protocol.

### DIFF
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -13,6 +13,7 @@ import {
     CollateralPosition,
     UndercollateralizedPosition,
     ExtrinsicData,
+    AccruedRewards,
 } from "../types";
 import { ApiPromise } from "@polkadot/api";
 import Big, { RoundingMode } from "big.js";
@@ -35,6 +36,7 @@ import {
 } from "../utils";
 import { InterbtcPrimitivesCurrencyId, LoansMarket } from "@polkadot/types/lookup";
 import { OracleAPI } from "./oracle";
+import { CurrencyId } from "../interfaces";
 
 /**
  * @category Lending protocol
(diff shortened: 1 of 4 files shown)


src/parachain/loans.ts
```diff
@@ -675,13 +677,120 @@ export class DefaultLoansAPI implements LoansAPI {
...
+            // Total KINT / INTR tokens awarded per block to suppliers of this market
+            this.api.query.loans.rewardSupplySpeed(underlyingCurrencyId),
+            this.api.query.tokens.totalIssuance(lendTokenId),
+        ]);
+        const lastIndex = Big(marketSupplyState.index.toString());
+        const supplyRewardSpeed = Big(marketSupplySpeed.toString());
+        const totalSupply = Big(totalIssuance.toString());
+
+        const deltaBlocks = currentBlockNumber - marketSupplyState.block.toNumber();
+        const deltaIndex = supplyRewardSpeed.div(totalSupply).mul(deltaBlocks);
```
daniel-savu, line 696: If the `totalSupply` is very large, the division might be rounded to zero. You should first multiply and then divide (in the other function as well)
```suggestion
        const deltaIndex = supplyRewardSpeed.mul(deltaBlocks).div(totalSupply);
```

peterslany: Great suggestion, thanks!

test/integration/parachain/staging/sequential/loans.test.ts
```diff
@@ -354,8 +353,45 @@ describe("Loans", () => {
...
+        it("should return correct amount of rewards", async () => {
+            await submitExtrinsic(
+                userInterBtcAPI,
+                await userInterBtcAPI.loans.lend(underlyingCurrency, newMonetaryAmount(1, underlyingCurrency, true)),
+                false
+            );
+
+            const rewards = await userInterBtcAPI.loans.getAccruedRewardsOfAccount(userAccountId);
+
+            expect(rewards.total.toBig().eq(1)).to.be.true;
```
daniel-savu, line 394: You can also lock the collateral and borrow `underlyingCurrencyId2` in the same test, to ensure the borrow reward is correctly calculated as well

peterslany: Good idea, thanks)

PR #638, opened 2023-05-23
add bitcoin tx / merkle proof parsing

gregdhill:
Didn't have time to finish but pushing the skeleton code in case anyone wants to pick it up while I'm OOO.


src/parachain/amm.ts
```diff
@@ -783,7 +783,7 @@ export class DefaultAMMAPI implements AMMAPI {
         }
 
         const lpTokenCurrencyId = newCurrencyId(this.api, pool.lpToken);
-        const farmDepositExtrinsic = this.api.tx.farming.deposit(lpTokenCurrencyId);
+        const farmDepositExtrinsic = this.api.tx.farming.deposit(lpTokenCurrencyId, 123);
```
peterslany, line 786: @sander2 @gregdhill I assume this was just a quick fix to silent the TSC typecheck. Regarding this change, I noticed that new parameter is `length_rewards: upper bound for number of reward currencies`, how do we need to handle this on lib side?

test/integration/parachain/staging/sequential/replace.test.ts
```diff
@@ -123,11 +123,13 @@ describe("replace", () => {
                     // eslint-disable-next-line no-constant-condition
                     while (true) {
                         const header = await api.rpc.chain.getHeader(hash);
-                        try {
-                            hash = await api.rpc.chain.getBlockHash(header.number.toNumber() + 1);
-                        } catch (_) {
-                            sleep(SLEEP_TIME_MS);
+                        const nextHash = await api.rpc.chain.getBlockHash(header.number.toNumber() + 1);
```
peterslany, line 126: This test had to be refactored because it started to fail: `api.rpc.chain.getBlockHash` started to return 0 instead if throwing when provided block height is higher than current block.

scripts/xcm-return-unknown-tokens.ts
```diff
@@ -1,84 +1,84 @@
...
-    const xcmTx = api.tx.xTokens.transferMultiasset(asset, dest, "Unlimited");
-
-    console.log("Constructed the tx, broadcasting first...");
-    await transactionAPI.sendLogged(setupTx, undefined);
-    console.log("broadcasting second...");
-    await transactionAPI.sendLogged(xcmTx, undefined);
-
-    api.disconnect();
-}
+// /* eslint @typescript-eslint/no-var-requires: "off" */
```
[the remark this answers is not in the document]
sander2: I think this screw prevented building the project, but if that's still an issue then we should probably fix that

src/parachain/amm.ts
sander2: Correct. You should set it to the length of the `farmingRewards.RewardCurrencies(lpTokenCurrencyId)`

src/types/bitcoin.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export type BitcoinNetwork = "mainnet" | "testnet" | "regtest";
+
+export type BlockHeader = {
+    merkleRoot: Uint8Array;
+    target: Uint8Array; // u256
+    timestamp: number;
+    version: number;
+    hash_: Uint8Array;
+    hashPrevBlock: Uint8Array;
```
sander2, line 11: We are using `Uint8Array`, `string` and `HexString`, which afaik are all representing the same thing: an array of bytes. For consistency, can we refactor to use only a single type?

sander2 (commented):
mostly lgtm, 1 suggestion for refactoring

test/unit/utils/bitcoin.test.ts
```diff
@@ -115,4 +117,64 @@ describe("Bitcoin", () => {
             assert.equal(encodedAddress, "bc1q75f6dv4q8ug7zhujrsp5t0hzf33lllnr3fe7e2pra3v24mzl8rrqtp3qul");
         });
     });
+
+    describe("getTxProof", () => {
```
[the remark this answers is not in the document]
sander2: Maybe try to hit a couple of code paths that are currently missed: coinbase transactions & different `TransactionLocktime` types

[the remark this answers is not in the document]
peterslany: Nice suggestions, added. Thanks)

src/types/bitcoin.ts
peterslany: Very good suggestion, I changed the types to use uniformly `HexString`. I would prefer `Buffer` as that's native type, but it was not possible because polkadotjs cannot convert it properly for some cases of witness data.

src/parachain/amm.ts
```diff
@@ -901,7 +902,12 @@ export class DefaultAMMAPI implements AMMAPI {
         }
 
         const lpTokenCurrencyId = newCurrencyId(this.api, pool.lpToken);
-        const farmWithdrawalExtrinsic = this.api.tx.farming.withdraw(lpTokenCurrencyId, amount.toString(true));
+        const rewardCurrencies = await this.api.query.farmingRewards.rewardCurrencies(lpTokenCurrencyId);
+        const farmWithdrawalExtrinsic = this.api.tx.farming.withdraw(
+            lpTokenCurrencyId,
+            amount.toString(true),
+            rewardCurrencies.size
```
sander2, line 909: Are you sure this will return the number of items, and not the size in bytes or something?

src/types/bitcoin.ts
```diff
@@ -0,0 +1,60 @@
...
+    inputs: Array<TransactionInput>;
+    outputs: Array<TransactionOutput>;
+    lockAt: TransactionLocktime;
+};
+
+export type TxStatus = {
+    confirmed: boolean;
+    confirmations: number;
+    blockHeight?: number;
+    blockHash?: string;
```
sander2, line 59: Shouldn't this be HexString as well? Not a blocker though

sander2 (approved):

src/parachain/amm.ts
peterslany: Yes. >@returns — the number of (unique) elements in Set.

src/types/bitcoin.ts
peterslany: I would like to, but it's not possible to do it in any meaningful way unfortunately because the conversion to hexadecimal string is done via polkadotjs converter that does use `string` as its return type, so we would need to do typecasting everywhere.

PR #640, opened 2023-05-26
chore: update docs generator and dev dependencies

ns212:
- Add dependabot for NPM packages
- Update dev dependencies
- Fix typescript error in `src/parachain/transaction.ts` which appeared after updating to TS 5.
- Clean-up `generate_docs` shell script
- Update the github action to build the docs for every commit but publish them only when running from master
- Bump nodejs to v18 for builds

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
             node-version: "18"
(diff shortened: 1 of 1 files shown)


peterslany (changes requested):
@ns212 Looks good, but can you remove the `package-lock.json` and instead use `yarn` to install new dependencies please?

PR #642, opened 2023-06-07
fix(loans): make subsidy reward index same as in the parachain

peterslany:
Fixes accrued reward calculation to match the decimals used on parachain side.
Solves part of the https://github.com/interlay/interbtc-ui/issues/1265

### DIFF
--- a/src/parachain/loans.ts
+++ b/src/parachain/loans.ts
@@ -701,7 +701,8 @@ export class DefaultLoansAPI implements LoansAPI {
     async _getLatestSupplyIndex(
         underlyingCurrencyId: CurrencyId,
         lendTokenId: CurrencyId,
-        currentBlockNumber: number
+        currentBlockNumber: number,
+        rewardCurrency: CurrencyExt
     ): Promise<Big> {
         const [marketSupplyState, marketSupplySpeed, totalIssuance] = await Promise.all([
             this.api.query.loans.rewardSupplyState(underlyingCurrencyId),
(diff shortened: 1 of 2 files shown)


bvotteler (approved):

src/parachain/loans.ts
```diff
@@ -730,27 +737,31 @@ export class DefaultLoansAPI implements LoansAPI {
...
             this.api.query.loans.rewardSupplierIndex(underlyingCurrencyId, accountId),
             this.api.query.tokens.accounts(accountId, lendTokenId),
         ]);
-        const supplierIndex = decodeFixedPointType(rewardSupplierIndex);
+        const supplierIndex = Big(rewardSupplierIndex.toString());
         const lendTokenBalance = Big(lendTokenRawBalance.free.toString()).add(lendTokenRawBalance.reserved.toString());
         const deltaIndex = latestSupplyIndex.sub(supplierIndex);
 
-        const accruedRewardInPlanck = deltaIndex.mul(lendTokenBalance);
+        const accruedRewardInPlanck = deltaIndex.mul(lendTokenBalance).div(10 ** rewardCurrency.decimals);
```
bvotteler, line 748: Not blocking: Is the variable name `accruedRewardInPlanck` still accurate?
It's not immediately clear to me if we're simply dividing by 10**decimals because the index now has been multiplied by it, or whether this normalizes the value.

peterslany: This is still accurate because index is multiplied by 10 ** decimals, so to compute the reward in Planck we have to divide it

PR #647, opened 2023-07-03
fix(amm): read farming reward curencies from farmingRewards pallet

peterslany:
Fixing farming reward currencies fetching for `amm.getClaimableFarmingRewards` in case the farming rewards are over, but there are still pending unclaimed rewards.

[Thread](https://discord.com/channels/745259537707040778/1123855798611759104/1123855953289293887)

### DIFF
--- a/src/parachain/amm.ts
+++ b/src/parachain/amm.ts
@@ -597,12 +597,9 @@ export class DefaultAMMAPI implements AMMAPI {
     private async _getFarmingRewardCurrencyIds(
         lpTokenCurrencyId: InterbtcPrimitivesCurrencyId
     ): Promise<Array<InterbtcPrimitivesCurrencyId>> {
-        const rewardCurrencies: Array<InterbtcPrimitivesCurrencyId> = [];
-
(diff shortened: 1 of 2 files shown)


src/parachain/amm.ts
```diff
@@ -594,22 +594,38 @@ export class DefaultAMMAPI implements AMMAPI {
...
 
+    private async _getFarmingRewardCurrencyIds(
+        lpTokenCurrencyId: InterbtcPrimitivesCurrencyId
+    ): Promise<Array<InterbtcPrimitivesCurrencyId>> {
+        const rewardCurrencies: Array<InterbtcPrimitivesCurrencyId> = [];
+
+        const rewardCurrenciesRaw = await this.api.query.farmingRewards.rewardCurrencies(lpTokenCurrencyId);
+        rewardCurrenciesRaw.forEach((rewardCurrencyId: InterbtcPrimitivesCurrencyId) =>
+            rewardCurrencies.push(rewardCurrencyId)
+        );
```
bvotteler, line 605: Suggestion (not blocking): Maybe something like this will work to turn the btreeset into an array directly?

```
const rewardCurrencies = Array.from(rewardCurrenciesRaw.values());
```

PR #664, opened 2023-07-28
feat: handle runtime upgrade 1.25

### LINKED ISSUE #663: Submit coinbase proofs for issue
To accompany [this fix](https://github.com/interlay/interbtc/pull/1112) (following https://github.com/interlay/interbtc-clients/issues/501) we need to relay two transactions (and merkle proofs) for each issue request using Esplora. The endpoint `/block/:hash/txids` will return all txids in a block and then we can fetch the **first** transaction using `/tx/:txid` - see [this comment](https://github.com/interlay/interbtc/pull/1112#issuecomment-1639183274). To verify that is the coinbase transaction we can check there is only one transaction input and `is_coinbase` is true.

To be more specific, [`getTxProof`](https://github.com/interlay/interbtc-api/blob/5324300a87774adef5aac5d89691e6e4852c8d56/src/utils/bitcoin.ts#L140) should return [`FullTransactionProof`](https://github.com/interlay/interbtc/blob/35507494078d8c168efc5bd9fd115fd9be0f6cbb/crates/bitcoin/src/types.rs#L31) where `user_tx_proof` and `coinbase_proof` are supplied in the current format ([`PartialTransactionProof`](https://github.com/interlay/interbtc/blob/35507494078d8c168efc5bd9fd115fd9be0f6cbb/crates/bitcoin/src/merkle.rs#L26)).

I also noticed we still allow execution of redeem and replace requests. We don't necessarily need to remove those methods if we change the function above but I also don't think they are used in the front-end.

peterslany:
Updating parachain to `1.25.0` version.

- Coinbase tx proof for user transaction is now provided for issue, redeem, replace requests execution
- [BREAKING]: `oracle.isOnline` now takes parameter of currency for which to check online status, universal oracle status is removed
- [BREAKING]  `system.getStatusCode` was removed, there is no more global parachain status, invalid tx submission should be prevented on the UI side by dry-running

Closes #663


src/external/electrs.ts
```diff
@@ -422,6 +429,16 @@ export class DefaultElectrsAPI implements ElectrsAPI {
         return this.getData(this.txApi.getTxHex(txid));
     }
 
+    async getCoinbaseTxId(userTxId: string): Promise<string | undefined> {
+        const blockHash = (await this.getTxStatus(userTxId)).block_hash;
+        console.log("utxid", userTxId, "blockhash", blockHash);
```
bvotteler, line 434: Do we still need the console.log here?

src/utils/bitcoin.ts
```diff
@@ -174,7 +174,34 @@ export async function getTxProof(
...
+    electrsAPI: ElectrsAPI,
+    userTxId: string
+): Promise<{
+    userTxProof: PartialTxProof;
+    coinbaseProof: PartialTxProof;
+}> {
+    const coinbaseTxId = await electrsAPI.getCoinbaseTxId(userTxId);
+
+    if (coinbaseTxId === undefined) {
+        throw new Error(`Coinbase txid not found for transaction ${userTxId}`);
```
bvotteler, line 191: Just for clarification: Is this going to be a problem if I send my BTC from a non-coinbase wallet?

test/unit/utils/bitcoin.test.ts
```diff
@@ -120,20 +120,22 @@ describe("Bitcoin", () => {
...
         const mockElectrsGetParsedExecutionParameters = (merkleProofHex: string, txHex: string) => {
+            const stubbedElectrsApi = sinon.createStubInstance(DefaultElectrsAPI);
             const [proof, tx] = [BitcoinMerkleProof.fromHex(merkleProofHex), bitcoinjs.Transaction.fromHex(txHex)];
             stubbedElectrsApi.getParsedExecutionParameters.withArgs(sinon.match.any).resolves([proof, tx]);
+            stubbedElectrsApi.getCoinbaseTxId.withArgs(sinon.match.any).resolves(tx.getId());
+            return stubbedElectrsApi;
         };
 
-        it.only("should parse proof and transactions correctly", async () => {
+        it("should parse proof and transactions correctly", async () => {
```
bvotteler, line 131: Woops, nice catch.

bvotteler (approved):
LGTM, no blockers, just a few hints / questions for clarification

src/external/electrs.ts
peterslany: nice catch!

src/utils/bitcoin.ts
peterslany: Coinbase wallet as the wallet developed by that CEX company? No, actually each bitcoin block has[ coinbase transaction ](https://developer.bitcoin.org/reference/transactions.html#coinbase-input-the-input-of-the-first-transaction-in-a-block)which is the first tx in the block with reward for the miner.

bvotteler: Riiight, the name keeps confusing me :)

PR #672, opened 2023-08-15
Chore: Change lib to esnext module, upgrade to polkadot/api 10.9.1

### LINKED ISSUE #670: Chore: Upgrade polkadot-js api
We need to upgrade to version 10.3.1 or newer as part of the runtime upgrade to `1.25.0`
If possible go to the latest version.

bvotteler:
Resolves #670 

Update polkadot/api to latest (`10.9.1`) so it works with our newest runtime `1.25.0`.
Had to change module from previously 'CommonJs' to 'esnext' due to a blocking issue with polkadot typegen ( polkadot-js/api#5540 ) which only has a workaround for esnext modules.

TODO before ready for review:
- [x] migrate unit tests from mocha -> jest
- [x] migrate integration tests from mocha -> jest
- [x] unit tests run and pass
- [x] integration tests run and pass
- [x] check if we want to consider another workaround and add manual definitions for those that cannot be generated until upgraded (did that and provided [lib version 2.4.3](https://github.com/interlay/interbtc-api/releases/tag/2.4.3))
- [x] revert manual definition for `transactionPaymentApi.queryInfo` (#673)


src/utils/bitcoin-core-client.ts
```diff
@@ -3,7 +3,9 @@
 import { MonetaryAmount } from "@interlay/monetary-js";
 import Big from "big.js";
 import { WrappedCurrency } from "../types";
+// import { createRequire } from "module";
 
+// const require = createRequire(import.meta.url);
```
peterslany, line 8: Is this needed to be commented out here?

peterslany (approved):
Good job handling the migration to jest :clap:

src/utils/bitcoin-core-client.ts
bvotteler: Good catch, not needed anymore. That is leftover from an earlier attempt to use `require` inside esm.

PR #697, opened 2023-10-09
Refactor: move BitcoinCoreClient to tests

bvotteler:
The unusual way we include the js-only library "bitcoin-core" via require causes issues when including the lib as dependency.

Since the BitcoinCoreClient is only used in integration tests anyway, I moved it out of the regular `src` tree and into the `test` tree. I also removed the test sources folder from the build ouput.


peterslany (approved):
LGTM

PR #724, opened 2024-02-09
add script to register asset

gregdhill:
The scripts no longer compile so I has to tweak the `tsconfig.json` to work with CommonJS so don't merge this.

Use `ts-node scripts/register-asset.ts`

### DIFF
--- a/scripts/register-asset.ts
+++ b/scripts/register-asset.ts
@@ -88,6 +88,7 @@ const NEW_ASSET = {
 function vaultRegistryCalls(
     api: ApiPromise,
     collateralCurrency: { ForeignAsset: number } | { LendToken: number },
+    vaultParams: any,
 ) {
     const {
         wrappedCurrency,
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -2,7 +2,6 @@
     "extends": "./tsconfig-base.json",
     "compilerOptions": {
         "target": "ES6",
-        "module": "ES6",
-        "outDir": "build/esm",
+        "module": "CommonJS",
```
sander2, line 5: Does this not cause issues for the ui? Fwiw I had this
```diff
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,8 +1,15 @@
 {
     "extends": "./tsconfig-base.json",
+    "ts-node": {
+        // these options are overrides used only by ts-node
+        // same as the --compilerOptions flag and the TS_NODE_COMPILER_OPTIONS environment variable
+        "compilerOptions": {
+          "module": "commonjs"
+        }
+      },
```

gregdhill: See pr description

sander2: I think my version is mergable but yea I guess we don't want the register-asset script

PR #727, opened 2024-02-23
Feat: enable vault withdraw all collateral

### LINKED ISSUE #725: Enable all collateral to be withdrawn
Blocks https://github.com/interlay/interbtc-ui/issues/1580

The UI calls the `vaults.withdrawCollateral` method which takes an amount with a type of `MonetaryAmount`. To withdraw all collateral, we need to be able to pass in `None` rather than a `MonetaryAmount`.

bvotteler:
Resolves #725 

Added new endpoints to easier withdraw all collateral for vaults.

**Note**: Tested manually. 
No integration test case sadly, because all the spun up and funded vaults end up with some tokens issued or to-be-issued left over from previous non-idempotent test cases.


sander2 (approved):
lgtm. Would it be useful to provide a `canWithdrawAll` or something for the UI? Or can UI easily detect that already?

sander2:
> No integration test case sadly, because all the spun up and funded vaults end up with some tokens issued or to-be-issued left over from previous non-idempotent test cases.
Could you not do a selfredeem? (there is a special tx for selfredeem that doesn't involve any btc payments)

bvotteler:
> Could you not do a selfredeem? (there is a special tx for selfredeem that doesn't involve any btc payments)
I tried a few angles with the existing test redeem utility code but had no luck. Or at least it didn't settle immediately.
I can give self redeem a try, I'd expect that requires the vault to get enough tokens issued from somewhere else, right?

bvotteler:
> lgtm. Would it be useful to provide a `canWithdrawAll` or something for the UI? Or can UI easily detect that already?

There are a few methods that should do the trick already. The two first that come to mind are: 
1. `vaultsAPI.getVaultCollateralization()` ([link](https://github.com/interlay/interbtc-api/blob/e01fc140fdc8b9019ab0f5ac7f2897776b0136bb/src/parachain/vaults.ts#L68-L87)) returns `undefined` if no tokens have been issued, or
2. Fetch vault by account id and collateral currency to get a `VaultExt` instance, then query for backed tokens which includes both issued and to-be-issued tokens:
```ts
const vaultExt = await vaultsApi.get(vaultAccountId, collateralCurrency));
const issuedAmount = vaultExt.getBackedTokens(); // includes to-be-issued
```
([Link to signature](https://github.com/interlay/interbtc-api/blob/e01fc140fdc8b9019ab0f5ac7f2897776b0136bb/src/parachain/vaults.ts#L61-L67) for get, and [link to getBackedTokens](https://github.com/interlay/interbtc-api/blob/e01fc140fdc8b9019ab0f5ac7f2897776b0136bb/src/types/vault.ts#L79-L81))

sander2:
> > Could you not do a selfredeem? (there is a special tx for selfredeem that doesn't involve any btc payments)
> 
> I tried a few angles with the existing test redeem utility code but had no luck. Or at least it didn't settle immediately. I can give self redeem a try, I'd expect that requires the vault to get enough tokens issued from somewhere else, right?

Yea that's right (but u could just sudo-mint some if you don't mind some state inconsistencies)

bvotteler:
> > > Could you not do a selfredeem? (there is a special tx for selfredeem that doesn't involve any btc payments)
> > 
> > 
> > I tried a few angles with the existing test redeem utility code but had no luck. Or at least it didn't settle immediately. I can give self redeem a try, I'd expect that requires the vault to get enough tokens issued from somewhere else, right?
> 
> Yea that's right (but u could just sudo-mint some if you don't mind some state inconsistencies)

Thanks for the self redeem tip! 
That helped in adding a test: commit 252682f

